PR #22, opened 2015-11-04
unit test improvements

corfor:
- corrected spelling of 'corporation'
- output serialized test results for easier visual inspection


corfor:
The most important feature for my usage is
- support absolute paths returned from IUrlPathBuilder

joukevandermaas:
I agree and I think I'll end up removing that else branch later. It's less confusing to just make prefixes explicit and move them into the url builder. The test improvements look great. Thanks!

PR #23, opened 2015-11-05
refactor CanonicalUrlPathBuilder

corfor:
It seems that CanonicalUrlPathBuilder was not finished? So I finished it, at least according to my understanding and needs.


joukevandermaas:
You're correct, I didn't push those changes yet but I did fully implement this class, among other things. I'll see if I can push those tonight.

Either way I can't merge this right now, because you (accidentally) added a `.orig` file.

PR #33, opened 2015-11-09
bug fix - missing Json Api media type

corfor:
Serious bug. Media type not registered, so NO json api support at all when used in web site.


joukevandermaas:
Thanks! This is not the first time a bug like this has slipped through CI testing, so I'll spend some time to improve coverage across the board.

PR #34, opened 2015-11-10
bug fix - 'related' link missing for canonical HasMany relationship

corfor:


joukevandermaas:
This was a deliberate choice, since the url generation scheme is to generate the absolute path (that you would write up if you knew nothing about the relationship, just the related resource) to a resource. You can't do that for `hasMany` relationships, because unlike resources with an id they don't have two endpoints referring to the same resource.

Example:
* `/people/123/job` refers to the same unique resource as `/companies/456`.
* `/people/123/friends` does not refer to the same unique resource as `/companies`.

Can you give some motivation for the need to have a different, non-similar looking url generated for the second case that is not covered in the `DefaultUrlPathBuilder`? Do you feel like this 'missing' link can be explained in documentation instead of including it?

PR #37, opened 2015-11-12
fix bug: GetPerRequestFormatterInstance not passing along JsonConverters

corfor:


joukevandermaas:
Looks good. I'll write an integration test for this later this week.

PR #40, opened 2015-11-15
Sorting support

### LINKED ISSUE #16: sorting support
Similar to #12. The implementation can probably be similar to #15 as well, except that sorting order should probably be decided by the client.

joukevandermaas:
Todo:

- [x] Add support in `JsonApiSerializer<T>`
- [x] `IEnumerable` support
- [x] Refactor pagination to use sorting interpreter internally
- [x] Add attribute to allow user queries
- [ ] Write documentation
- [ ] Add more tests
  - [x] Tests for useful exceptions (e.g. sort on property that does not exist, do something unsupported, etc.)
  - [ ] Unit tests of components
  - [x] Test that it works together well with pagination
  - [x] Integration tests for media type formatter

closes #16

### DIFF
--- a/Tests/Controllers/CompaniesController.cs
+++ b/Tests/Controllers/CompaniesController.cs
@@ -10,19 +10,21 @@ namespace Tests.Controllers
     public class CompaniesController : ApiController
     {
         [HttpGet]
+        [Route("companies/{id}")]
         public Company GetCompany(string id)
         {
             return new Company(prefill: true);
         }
 
         [HttpGet]
-        [Paginated]
+        [Paginated(PerPage = 12)]
+        [Route("companies")]
         public IEnumerable<Company> GetCompanies()
         {
             return GetEnumerable().Take(100);
         }
 
-        private IEnumerable<Company> GetEnumerable()
+        private static IEnumerable<Company> GetEnumerable()
         {
             var i = 0;
             while (true)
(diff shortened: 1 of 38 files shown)


Tests/Controllers/CompaniesController.cs
```diff
@@ -13,9 +10,27 @@ namespace Tests.Controllers
...
+
+        [HttpGet]
+        [Paginated(PerPage = 12)]
+        [Route("companies")]
+        public IEnumerable<Company> GetCompanies()
+        {
+            return GetEnumerable().Take(100);
+        }
+
+        private static IEnumerable<Company> GetEnumerable()
```
joukevandermaas: This should probably be a method on a static helper class, since now it's duplicated a lot.

Tests/Controllers/PeopleController.cs
```diff
@@ -11,29 +12,44 @@ namespace Tests.Controllers
...
         }
 
+        [HttpGet]
+        [Route("people")]
+        public IEnumerable<Person> GetPeople()
+        {
+            return GetPersonEnumerable().Take(100);
+        }
+
         private static IEnumerable<Person> GetPersonEnumerable()
```
joukevandermaas: Same as above

joukevandermaas:
I'm merging this to create a pre-release for early testing. See #41 for what is left in terms of testing.

PR #44, opened 2015-11-19
add recursive exception message to ApiError

corfor:
Meaningful errors were being suppressed when, for example, I accidentally added a duplicate Attribute in my Resource definition.


joukevandermaas:
I'm not 100% about the use of space as a separator. If an exception message does not end in a period it may look weird. I propose you call `EnsureEndsWith(".")` on the message or separate with `\n`. 

Otherwise it looks great! Thanks!

corfor:
The suggestions you made have been implemented.

Issue #61, opened 2015-12-22
Nested includes and compound documents


jrexdepew:
Currently in one of our projects we have a resource that we include all of its relationships to prevent multiple api calls. One of the relationships of the parent resource also has has relationships we would like to include all in one payload. For example let's say I have bars resource that has many beers and each beer has a manufacturer.  In my payload I would like to include all the beers a bar sells and then include the manufacturers inside each beer object.  After running through the code it looks like when processing relationship resources, saule is not also processing that relationships resources recursively even if the object being serialized includes that information. I suspect this may have been due to circular references?  If this is possible and I have something setup incorrectly then let me know.  I am currently on the commute home but when I get home I can provide a code example if needed.  I believe if this isn't currently possible, the method that processes the top level resource in the serializer could be refactored to be called recursively for each included resource.  Please excuse any typos as I am typing this on a phone :(

joukevandermaas:
You are exactly right. When first implementing this I decided not to add this feature for fear of infinite loops.

However I'm pretty sure Saule already checks for circular references, so this should be relatively straightforward to implement.

rhyek:
Hi. Is this feature available in the latest beta NuGet package (saule.1.5.0.305-beta)? Currently running some tests and I'm not getting the nested includes. Thanks.

rhyek:
I just built from source and it's definitely not working. My models look like this:
```
namespace api.Models.Db {
    public class RefreshToken {
        [Key]
        public string Id { get; set; }
        [Required]
        [MaxLength(100)]
        public string Subject { get; set; }
        [Required]
        [MaxLength(50)]
        public string Client { get; set; }
        public DateTimeOffset Issued { get; set; }
        public DateTimeOffset Expires { get; set; }
        [Required]
        public string ProtectedTicket { get; set; }
        public virtual ConnectionSource ConnectionSource { get; set; }
    }
    public class ConnectionSource {
        [Key]
        public string Id { get; set; }
        public string OS { get; set; }
        public string UserAgent { get; set; }
        public string Device { get; set; }
        public string RemoteIp { get; set; }
    }
}

namespace api.Models.JsonApi {
    public class RefreshTokenResource : ApiResource {
        public RefreshTokenResource() {
            Attribute("Client");
            Attribute("Issued");
            Attribute("Expires");
            BelongsTo<ConnectionSourceResource>("ConnectionSource");
        }
    }
    public class ConnectionSourceResource : ApiResource {
[...]

PR #68, opened 2016-01-03
Allow models to have recursive references

### LINKED ISSUE #67: Self referencing loop detected
I ran into a problem with my resources that reference each other, as seen below.

```
// ApplicationResource
BelongsTo<LevelResource>("Level");

// LevelResource
HasMany<ApplicationResource>("Applications");
```

The API is returning the data without an issue, but I get the following InvalidOperationException when an attempt is made to serialize the results:

```
"The 'ObjectContent`1' type failed to serialize the response body for content type 'application/vnd.api+json'. Self referencing loop detected with type 'Models.Application'. Path 'Level.Applications'."
```

If I remove the Applications collection from the Level object the results are serialized as expected. Prior to using Saule, I got around this issue by setting the `ReferenceLoopHandling` property on `JsonFormatter.SerializerSettings` to `ReferenceLoopHandling.Ignore`.

I wasn't sure if this was related to #18 or if it's something else.

joukevandermaas:
fixes #67

### DIFF
--- a/Saule/JsonApiSerializer.cs
+++ b/Saule/JsonApiSerializer.cs
@@ -88,7 +88,7 @@ namespace Saule
 
         private JsonSerializer GetJsonSerializer()
         {
-            var serializer = new JsonSerializer();
+            var serializer = new JsonSerializer { ReferenceLoopHandling = ReferenceLoopHandling.Ignore };
             foreach (var converter in JsonConverters)
             {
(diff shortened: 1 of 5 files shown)


joukevandermaas:
Note that this may conflict with the implementation of #61 in some cases, but I doubt it'll be an issue.

Issue #84, opened 2016-01-20
HttpError is suppressed


corfor:
When accessing an endpoint that is decorated with [Authorize] but the user is not authenticated, the misleading error is returned:
```
{
  "errors": [
    {
      "title": "You must add a [ReturnsResourceAttribute] to action methods.",
      "detail": "Saule.JsonApiException: You must add a [ReturnsResourceAttribute] to action methods.",
      "code": "Saule.JsonApiException",
      "links": {
        "about": "https://github.com/joukevandermaas/saule/wiki"
      }
    }
  ]
}
```

But really all that happened is an HttpError was raised indicating `401 Not Authorized`. Would be better to allow underlying HttpError to surface.
```
{
  "errors": [
    {
      "title": "Authorization has been denied for this request."
    }
  ]
}
```

joukevandermaas:
I pushed a new pre-release (just now, can take some time to appear) and I'll also create a new 1.4 release with a fix for this bug.

Thanks for reporting!

PR #85, opened 2016-01-20
Bugfix for HttpError is suppressed #84

corfor:


joukevandermaas:
I did not notice this PR before making #86, but since this one has a build error I'll just merge the other.

Thanks anyway for submitting and sorry for missing it.

PR #93, opened 2016-04-14
[WIP] error renderers for different exception types

NullVoxPopuli:
Currently, this doesn't work.

I guess I'm having a hard time figuring out where exactly the exceptions are caught and rendered as json


Saule/Saule.csproj
```diff
@@ -41,6 +41,14 @@
     <CodeAnalysisIgnoreGeneratedCode>true</CodeAnalysisIgnoreGeneratedCode>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="EntityFramework, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089, processorArchitecture=MSIL">
```
joukevandermaas: We definitely don't want to pull in Entity Framework as a reference. Why do you think we need this?

Saule/Saule.nuspec
```diff
@@ -14,4 +14,9 @@
     <copyright>Copyright 2015</copyright>
     <tags>web api serialization media formatter json api jsonapi ember data</tags>
   </metadata>
+  <files>
+    <file src="..\Saule\bin\Debug\Saule.dll" target="lib\net46\"></file>
```
joukevandermaas: Is there a good reason to include the debug build in the nuget package?

Saule/Serialization/ErrorRenderers/DbEntityValidationExceptionRenderer.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Collections.Generic;
+using System.Data.Entity.Validation;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Newtonsoft.Json.Linq;
+
+namespace Saule.Serialization.ErrorRenderers
+{
+    internal class DbEntityValidationExceptionRenderer : ExceptionRenderer
```
joukevandermaas: As said above, I think this belongs in people's projects, not Saule. We don't want to depend on entity framework.

Saule/Serialization/ApiError.cs
```diff
@@ -50,6 +73,26 @@ private static string GetRecursiveExceptionMessage(HttpError ex)
             }
 
             return msg;
+        }
+
+        private ExceptionRenderer GetRenderer()
```
joukevandermaas: Let's think of a more flexible way to pick a renderer. We'd want people to supply one somehow. Perhaps using a factory in the configuration object?

Saule/Saule.csproj
NullVoxPopuli: For thee executions?

Saule/Saule.nuspec
NullVoxPopuli: Im just debugging locally

Saule/Serialization/ApiError.cs
NullVoxPopuli: I imagine one would need too be supplied at runtime, otherwise, people would need to define their own GetRenderer

NullVoxPopuli: though, I don't even know if this is the right approach, because my error received in Postman is the same as it was before all these changes

joukevandermaas:
What is the status of this PR?

NullVoxPopuli:
I think there need to be some decisions around how extensions to Saule happen -- how to determine what renders an Exception, etc. It doesn't make sense to depend on EF in Saule in order to render EF errors, but it should be super straight forward to write on EFErrorRenderer or something like that, ya know?

joukevandermaas:
I agree, there should be an extension point for this. It should likely be some interface people can implement and pass to Saule using the configuration object (similar to the URL builder). The interface should probably look something like this:

```csharp
public interface ErrorRenderer 
{
    IEnumerable<JsonApiError> Render(object error);
}
```

where `JsonApiError` is a class that has all the error fields as [described](http://jsonapi.org/format/#error-objects) in the spec.

NullVoxPopuli:
I like that. How would saule know to use a custom renderer?

joukevandermaas:
Same as the URL builder; if the consumer supplies one, Saule will use it. When people build their own, they can derive from a default implementation and only override what they are actually customizing.

joukevandermaas:
@NullVoxPopuli I'm closing this for now. Feel free to open a new PR with the changes we discussed against the current `master` branch.

PR #96, opened 2016-04-24
use Guids as id's

eltjo-k:
The use of Guids as id's leads to the following types of exception when id's are converted to strings:
```
{System.InvalidCastException: Object must implement IConvertible.
   at System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
   at Newtonsoft.Json.Linq.Extensions.Convert[T,U](T token)
   at Newtonsoft.Json.Linq.Extensions.Value[T,U](IEnumerable`1 value)
   at Newtonsoft.Json.Linq.Extensions.Value[U](IEnumerable`1 value)
   at Saule.Serialization.ResourceSerializer.<>c__DisplayClass23_0.<GetRelationshipData>b__0(IDictionary`2 props)
   at Saule.Serialization.ResourceSerializer.SerializeArrayOrObject(JToken token, Func`2 serializeObj)
   at Saule.Serialization.ResourceSerializer.GetRelationshipData(ResourceRelationship relationship, JToken relationshipValues)
   at Saule.Serialization.ResourceSerializer.SerializeRelationship(ResourceRelationship relationship, IDictionary`2 properties)
   at Saule.Serialization.ResourceSerializer.SerializeRelationships(IDictionary`2 properties)
   at Saule.Serialization.ResourceSerializer.SerializeData(IDictionary`2 properties)
   at Saule.Serialization.ResourceSerializer.SerializeArrayOrObject(JToken token, Func`2 serializeObj)
   at Saule.Serialization.ResourceSerializer.Serialize(JsonSerializer serializer)
   at Saule.JsonApiSerializer.Serialize(PreprocessResult result)
   at Saule.JsonApiSerializer`1.Serialize(Object object, Uri requestUri)
   at ...
```

This happens with:
- collections of objects with Guid-ids
- single objects with `belongsTo`-relations to objects with Guid-ids
- single objects with `hasMany `-relations to objects with Guid-ids.

Unit tests for these cases have been added.

These exceptions can be solved by replacing occurences of `JToken.Value<string>()` by cast operators that are explicitly defined for `JToken`'s (see [here](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Linq/JToken.cs#L1223)). This cast operator throws exceptions if the `JToken` does not contain the proper datatypes as defined in [this list](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Linq/JToken.cs#L74), which seems to contain every type that might be used as an index. Finally it uses `Convert.ToString(...)` which first tries to convert through `IConvertible`, just like the `JToken.Value<string>()`-conversion method does, so this change should not break any existing working behaviour.

### DIFF
--- a/Saule/Serialization/ResourceSerializer.cs
+++ b/Saule/Serialization/ResourceSerializer.cs
@@ -183,7 +183,7 @@ namespace Saule.Serialization
                 data["links"] = AddUrl(
                     new JObject(),
                     "self",
-                    _urlBuilder.BuildCanonicalPath(_resource, EnsureHasId(properties, _resource).Value<string>()));
+                    _urlBuilder.BuildCanonicalPath(_resource, (string)EnsureHasId(properties, _resource)));
             }
 
             return data;
@@ -221,7 +221,7 @@ namespace Saule.Serialization
             var relToken = GetMinimumRelationship(
                 objId.ToString(),
                 relationship,
-                relationshipProperties != null ? GetId(relationshipProperties, relationship.RelatedResource).Value<string>() : null);
+                relationshipProperties != null ? (string)GetId(relationshipProperties, relationship.RelatedResource) : null);
             if (relationshipValues == null)
             {
                 return relToken;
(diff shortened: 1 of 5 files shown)


joukevandermaas:
Awesome work, thanks! I will create 1.4.2 with this fix in it.

PR #98, opened 2016-05-04
allow manual serialization

NullVoxPopuli:
I was trying to manually serialize, because I am writing a general CRUD controller.

basically, this is what I was trying to do:

```csharp
    public class DefaultCrudController<TDbType, TResource>: ApplicationController where TResource : Saule.ApiResource, new()
    {

        [HttpGet]
        [ReturnsResource(typeof(ApplicationResource))]
        public JToken Index()
        {
            var collection = CollectionFromUser();
            return RenderResource(collection);
         }

        [HttpGet]
        public JToken Show(string id)
        {
            var resource = Resource(id);
            return RenderResource(resource);
        }

        [HttpPut]
        public JToken Update(string id, [FromBody] JObject bodyParams)
        {
            var resource = Resource(id);
            resource.Update(bodyParams);
            return RenderResource(resource);
        }

        [HttpPost]
        public JToken Create([FromBody] JObject bodyParams)
        {
            var collection = CollectionFromUser();
            var obj = collection.Create(bodyParams);

            return RenderResource(obj);
        }

        [HttpDelete]
        public JToken Destroy(object id)
        {
            var resource = Resource(id);

            resource.Destroy();
            return RenderResponseForJson();
        }


        #region Helpers

        [NonAction]
        protected dynamic CollectionFromUser()
        {
            var serializer = new JsonApiSerializer<TResource>();
            var methodName = typeof(TDbType).Name.Pluralize(inputIsKnownToBeSingular: false);
            var collection = TypedCurrentUser.Send(methodName);
            return collection;
        }


        [NonAction]
        protected dynamic Resource(object id)
        {
            var obj = CollectionFromUser().Find(id);
            return obj;
        }

        [NonAction]
        protected JToken RenderResponseForJson(JToken json = null)
        {
            return null;
        }

        [NonAction]
        protected JToken RenderResource(dynamic obj)
        {
            var serializer = new JsonApiSerializer<TResource>();
            JToken json = serializer.Serialize(obj, Request.RequestUri);
            return json;
        }

        #endregion

    }

```

it's still a work in progress, but what was tripping me up was in here:

```csharp

        [NonAction]
        protected JToken RenderResource(dynamic obj)
        {
            var serializer = new JsonApiSerializer<TResource>();
            JToken json = serializer.Serialize(obj, Request.RequestUri);
            return json;
        }

```

it looks like Saule, no matter what, always tries to take the result of an action and tries to serialize it. 
That's great. But Since I can't use GenericTypes on `ReturnResource`, I had to come up with a different way to do this.

### DIFF
--- a/Saule/Http/JsonApiMediaTypeFormatter.cs
+++ b/Saule/Http/JsonApiMediaTypeFormatter.cs
@@ -110,6 +110,13 @@ namespace Saule.Http
                 preprocessed = PreprocessingDelegatingHandler.PreprocessRequest(value, _request, _config);
             }
 
+            var preSerializedJson = preprocessed.ResourceSerializer.Value as JObject;
+            if (preSerializedJson != null)
+            {
+                await WriteJsonToStream(preSerializedJson, writeStream);
+                return;
+            }
+
             var json = JsonApiSerializer.Serialize(preprocessed);
             await WriteJsonToStream(json, writeStream);
         }
--- a/Saule/Http/PreprocessingDelegatingHandler.cs
+++ b/Saule/Http/PreprocessingDelegatingHandler.cs
@@ -62,6 +62,11 @@ namespace Saule.Http
 
             var value = result.Content as ObjectContent;
 
+            if (value?.Value is Newtonsoft.Json.Linq.JObject)
+            {
+                return result;
+            }
+
             var content = PreprocessRequest(value?.Value, request, _config);
 
             if (content.ErrorContent != null)
--- a/Saule/Serialization/ResourceSerializer.cs
+++ b/Saule/Serialization/ResourceSerializer.cs
@@ -17,6 +17,8 @@ namespace Saule.Serialization
         private readonly IUrlPathBuilder _urlBuilder;
         private bool _isCollection;
 
+        internal object Value => _value;
+
         public ResourceSerializer(
             object value,
             ApiResource type,


Saule/Http/JsonApiMediaTypeFormatter.cs
```diff
@@ -110,6 +110,13 @@ public override bool CanWriteType(Type type)
                 preprocessed = PreprocessingDelegatingHandler.PreprocessRequest(value, _request, _config);
             }
 
+            var preSerializedJson = preprocessed.ResourceSerializer.Value as JObject;
```
joukevandermaas: Maybe this code should be in the `PreprocessingDelegatingHandler`? That way it also works from `JsonApiSerializer`. I know in your specific use-case that doesn't make any difference, but I've tried to keep the serializer and formatter in sync feature-wise.

Saule/Http/PreprocessingDelegatingHandler.cs
```diff
@@ -62,6 +62,11 @@ public PreprocessingDelegatingHandler(JsonApiConfiguration config)
 
             var value = result.Content as ObjectContent;
 
+            if (value?.Value is Newtonsoft.Json.Linq.JObject)
```
joukevandermaas: Shouldn't this be sufficient to just pass through json responses? I'm not sure why the changes to the formatter are needed.

joukevandermaas:
Great work! Thanks for contributing.

It looks like your build failed due to a style cop error. You can see those locally by doing a release build.  Also, please add some tests for this behavior so we don't accidentally break this feature later.

Saule/Http/PreprocessingDelegatingHandler.cs
NullVoxPopuli: good question. I'll try removing the changes in the other file, and see what happens

NullVoxPopuli: this is what happens if I comment out the other stuff:

```
{
  "errors": [
    {
      "title": "The 'ObjectContent`1' type failed to serialize the response body for content type 'application/vnd.api+json'. Resources must have an id.",
      "code": "System.InvalidOperationException"
    }
  ]
}
```

joukevandermaas:
@NullVoxPopuli I'm closing this PR for now because it has been sitting around for a while and is out of date with master. Feel free to open a new PR whenever.

PR #103, opened 2016-05-09
[WIP] Avoid object tree traversal

NullVoxPopuli:
Here is what I have so far.
I had to add a dep to one of my nugets that handles some implicit reflection stuff. (cause reflectively invoking properties / methods is really annoying)

### DIFF
--- a/Saule/Saule.csproj
+++ b/Saule/Saule.csproj
@@ -41,6 +41,9 @@
     <CodeAnalysisIgnoreGeneratedCode>true</CodeAnalysisIgnoreGeneratedCode>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="csharp-extensions">
+      <HintPath>D:\ClientDevelopment\AllisonTransmission\Integration.API\packages\csharp-extensions.1.2.5\lib\net46\csharp-extensions.dll</HintPath>
+    </Reference>
     <Reference Include="Humanizer, Version=1.37.7.0, Culture=neutral, PublicKeyToken=979442b78dfc278e, processorArchitecture=MSIL">
       <HintPath>..\packages\Humanizer.1.37.7\lib\portable-win+net40+sl50+wp8+wpa81+MonoAndroid10+MonoTouch10+Xamarin.iOS10\Humanizer.dll</HintPath>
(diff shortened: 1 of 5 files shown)


Saule/Saule.csproj
```diff
@@ -41,6 +41,9 @@
     <CodeAnalysisIgnoreGeneratedCode>true</CodeAnalysisIgnoreGeneratedCode>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="csharp-extensions">
+      <HintPath>D:\ClientDevelopment\AllisonTransmission\Integration.API\packages\csharp-extensions.1.2.5\lib\net46\csharp-extensions.dll</HintPath>
```
NullVoxPopuli: ha, this shouldn't be in here.

Saule/Serialization/ResourceSerializer.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using csharp_extensions.Extensions;
```
NullVoxPopuli: this is the dep

Saule/Serialization/ResourceSerializer.cs
```diff
@@ -66,71 +69,80 @@ public JObject Serialize(JsonSerializer serializer)
...
+        private static string EnsureHasId(object properties, ApiResource resource)
         {
             var id = GetId(properties, resource);
             if (id == null)
             {
                 throw new JsonApiException(ErrorType.Server, "Resources must have an id");
             }
 
-            return id;
+            return id.ToString();
```
NullVoxPopuli: It's better to do `.ToString()` here than to cast to string, cause of weird primary key types (like guids --  /cc @wozbo)

joukevandermaas:
@NullVoxPopuli it looks like you only added the reference locally, because it is not found on the build server. Do you think you'll be able to wrap up this PR? It looks pretty good so far.

NullVoxPopuli:
@joukevandermaas I don't think I'll be able to -- I'm no longer working with C# (switched jobs).
If you want, I can give you write permissions to my fork, though

PR #105, opened 2016-05-11
[WIP] try to add test for relationship data

NullVoxPopuli:
I'm not sure if this is the right place to put this. (the test also doesn't pass, cause I couldn't figure out how to not get a 500 error)

### DIFF
--- a/Tests/Controllers/PeopleController.cs
+++ b/Tests/Controllers/PeopleController.cs
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
 using System.Web.Http;
+using Newtonsoft.Json.Linq;
 using Saule.Http;
 using Tests.Helpers;
 using Tests.Models;
@@ -50,6 +51,13 @@ namespace Tests.Controllers
         {
             person.Identifier = id;
             return person;
+        }
+
+        [HttpPost]
+        [Route("relationships/people")]
+        public JObject PostParams([FromBody] JObject bodyParams)
+        {
+            return bodyParams;
         }
 
         [HttpGet]
--- a/Tests/Integration/ContentNegotiationTests.cs
+++ b/Tests/Integration/ContentNegotiationTests.cs
@@ -162,6 +162,37 @@ namespace Tests.Integration
 
                 Assert.Equal(HttpStatusCode.OK, result.StatusCode);
             }
+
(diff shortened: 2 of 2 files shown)


Tests/Controllers/PeopleController.cs
```diff
@@ -50,6 +51,13 @@ public Person PostPerson(string id, Person person)
         {
             person.Identifier = id;
             return person;
+        }
+
+        [HttpPost]
+        [Route("relationships/people")]
+        public JObject PostParams([FromBody] JObject bodyParams)
```
NullVoxPopuli: bodyParams should include a flattened version of the attributes and relationships, yeah?

currently, it only gives me attributes

joukevandermaas: We should probably add special case handling for deserializing into `JObject`. I'm not sure if the result should be:

1. Just pass through the request, untouched
2. Convert the request into a 'standard' web api request (bag of attributes, relationships and id).

I'm leaning towards 1, but I assume for your use-case 2 would be better?

NullVoxPopuli: yeah, the way ActiveModelSerializers does it, is that it flattens the attributes and relationships, and appropriately renames polymorphic keys,

```
relationships: {
   admin: {
      data: {
        id: 1, type: 'users'
      } 
   }
}
```

becomes:
```
admin_id: 1, admin_type: 'User'
```

joukevandermaas: Are you using `JObject` to mimic the dynamic nature of Ruby? If so, you might be better of just committing to type-safety in C#. Most (99%) of the time, you should be able to create a model class with the properties you want. Saule will already ignore properties that are not on the model.

NullVoxPopuli: I was actually using JObject, because we had to write a wrapper around EF, and our Create method takes either an anonymous object or a JObject to assign attributes. I could see if I could have it copy all the properties from the class. Right now my method signature is:

        public JToken Create([FromBody] TDbType bodyParams)

NullVoxPopuli: actually, looking at my instance of TDbType, it still doesn't have relationships.... :-\

joukevandermaas:
@NullVoxPopuli I'm closing this PR for the time being. Feel free to open a new PR if you want to continue working on this. Thanks!

PR #110, opened 2016-05-27
Serialize recursively placing all nested objects in the 'included' array

rhyek:
Added functionality that recursively serializes a model, placing all nested objects in the `included` array. This conforms to the JSON API spec, which states:

http://jsonapi.org/format/#document-top-level

A document MAY contain any of these top-level members:

- jsonapi: an object describing the server’s implementation
- links: a links object related to the primary data.
- **included: an array of resource objects that are related to the primary data and/or each other (“included resources”).**

### DIFF
--- a/Saule/Serialization/ResourceSerializer.cs
+++ b/Saule/Serialization/ResourceSerializer.cs
@@ -49,7 +49,7 @@ namespace Saule.Serialization
 
             var result = new JObject
             {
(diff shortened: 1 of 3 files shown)


joukevandermaas:
Thanks a lot for doing this! I will review your changes this weekend and hopefully publish a new pre-release.

Issue #116, opened 2016-06-14
PreprocessingDelegatingHandler breaks Swashbuckle embedded Swagger UI


nukefusion:
`PreprocessingDelegatingHandler` runs on every single request so requests for non-API resources (such as CSS and JS assets for the Swagger UI) are [rejected with a JsonApiException](https://github.com/joukevandermaas/saule/blob/5e162593848e22c44fef96358c10dba2a62b514f/Saule/Http/PreprocessingDelegatingHandler.cs#L47).

Is there any reason why the handler can't be executed only for requests with an accept header containing the JSON API media type?

joukevandermaas:
This was fixed in #118.

PR #118, opened 2016-06-15
Fix for #111 and #116

### LINKED ISSUE #111: 500 error when action returns a simple POCO object and using config.ConfigureJsonApi()
I'm using a local build with the latest source code. This is my action: 
```c#
[Authorize]
[Route("api/usuarios/{id}/parcial")]
public async Task<dynamic> PutPartial(string id, UsuarioPartialPut json) {
    if (id != User.Identity.GetUserId()) {
        throw new HttpResponseException(Request.CreateErrorResponse(HttpStatusCode.Forbidden, Errores.SinAcceso));
    }
    using (var db = new ApplicationDbContext()) {
        var user = await db.Users.Where(u => u.Id == id).SingleOrDefaultAsync();
        if (user == null) {
            return NotFound();
        }
        
        user.Email = json.Email;
        user.Locale = json.Locale;

        await db.SaveChangesAsync();
        return new {
            ok = true
        };
    }
}
```
In my `WebApiConfig.cs`, if I use `config.ConfigureJsonApi()` I will get a 500 http error, but I will still get the JSON payload: `{"ok":true}`.

If I use `config.Formatters.Add(new JsonApiMediaTypeFormatter())` I get the same payload, but no 500 status.

This action isn't meant to use Saule, and the request has header `Accept:application/json, text/javascript, */*; q=0.01`.

nukefusion:
This modification checks for the presence of the JSONAPI media type within the accept header before executing the body of the PreprocessingDelegatingHandler. This prevents the handler from breaking non-API routes for static or other content. An associated test has been added against a controller serving static content. Should resolve #111 and #116.

### DIFF
--- a/Saule/Http/PreprocessingDelegatingHandler.cs
+++ b/Saule/Http/PreprocessingDelegatingHandler.cs
@@ -11,6 +11,8 @@ using Saule.Serialization;
 
 namespace Saule.Http
 {
+    using System.Linq;
+
     /// <summary>
     /// Processes JSON API responses to enable filtering, pagination and sorting.
     /// </summary>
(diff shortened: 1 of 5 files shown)


joukevandermaas:
Thanks, this is awesome!

PR #124, opened 2016-07-23
Support belongsTo relationships when data is null

### LINKED ISSUE #123: Deserializing null relationships
Saule is fantastic! Thanks a million for all the work. Quick question:

My application is posting the following json to a Saule resource with a relationship and the deserializer is unable to deserialize the review object because its data is null. Any thoughts

```
{
  "data": {
    "attributes": {
      "name": null,
      "price": 12
    },
    "relationships": {
      "review": {
        "data": null
      }
    },
    "type": "book"
  }
}
```
The BookResource look like this:

```
public class BookResource : ApiResource
{
    public BookResource()
    {
        WithId("Id");
        Attribute("Name");
        Attribute("Price");
        HasMany<ReviewResource>("Review");
    }
}
```
I receive an "Object reference not set to an instance of an object." error during deserialization of the book's review. 

Is there a simple way to override the deserializer to a null check perhaps or am I maybe doing something blatantly wrong?

goo32:
Fixes #123

### DIFF
--- a/Saule/Serialization/ResourceDeserializer.cs
+++ b/Saule/Serialization/ResourceDeserializer.cs
@@ -27,7 +27,9 @@ namespace Saule.Serialization
             {
                 var obj = json["data"] as JObject;
 
-                if (obj == null) { return null; }
+                if (obj == null) {
+                    return null;
+                }
 
                 return SingleToFlatStructure(json["data"] as JObject);
             }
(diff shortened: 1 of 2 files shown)


goo32:
Would it be possible to cut a release that would include this fix?

joukevandermaas:
@goo32 I'll try to do that this week. In the meantime, the latest pre-release includes this fix.

goo32:
My apologies... I didn't realize there was a pre-release. I'll test my stuff out with the latest. Thanks!

joukevandermaas:
NP. Any commit on `master` publishes a new pre-release, so the latest pre-release is always up to date with whatever is in `master`.

PR #136, opened 2016-11-01
Related resources

yohanmishkin:
Here's a spike on #126. Let me know if you feel like it might be worth iterating on and polishing, but I used [JSON API Fetching Includes](http://jsonapi.org/format/#fetching-includes) for reference.

PR #137, opened 2017-01-28
Related resources

bjornharrtell:
Rebased from previous PR #136.


bjornharrtell:
Implementation seems to work in some limited tests. Some test cases was failing but I've now managed to fix them. @joukevandermaas this is now ready for review.

.gitattributes
```diff
@@ -1,15 +0,0 @@
-# Auto detect text files and perform LF normalization
```
joukevandermaas: What's the reason to remove this file? I think we still want it.

Saule/JsonApiSerializerOfT.cs
```diff
@@ -46,6 +47,11 @@ public JsonApiSerializer()
         public bool Paginate { get; set; } = false;
 
         /// <summary>
+        /// True if responses should be paginated, otherwise false.
```
joukevandermaas: Copy/paste error?

Tests/JsonApiSerializerTests.cs
```diff
@@ -162,6 +162,27 @@ public void AppliesSorting()
             Assert.Equal(expected, props);
         }
 
+        [Fact(DisplayName = "Applies including when includes specified")]
+        public void AppliesIncluding()
```
joukevandermaas: It could be good to have a test scenario like this:

1. two resources that both have a relationship to the same third resource
2. both include it
3. it's only in the included hash once

Saule/Queries/Including/IncludingInterpreter.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using System.Collections;
+using System.Linq;
+using Saule.Queries.Including;
+
+namespace Saule.Queries.Including
+{
+    internal class IncludingInterpreter
```
joukevandermaas: It doesn't look like this is needed/used (?)

.gitattributes
bjornharrtell: It caused git clone to autoconvert some files (no matter which OS or method it seems). I think @yohanmishkin had the same problem. Perhaps it can be adjusted rather than removed. The default .gitattributes that Visual Studio 2015 creates has only `* text=auto` so perhaps that would work and be the "correct" way to handle it.

Saule/JsonApiSerializerOfT.cs
bjornharrtell: Perhaps, but it (the property) also does not seem to be used and I can't see what it's purpose was intended for. Will remove it.

Saule/Queries/Including/IncludingInterpreter.cs
bjornharrtell: Agreed, removing.

bjornharrtell:
There are two behavioral changes that I've seen with this PR:

1. Before it seems the default behavior was to return all relationship data as included. With this PR relationship data will not be included by default and must be explicitly requested with the include param. I think this behavior was added intentionally by @yohanmishkin but I'm not sure about the reasoning.

2. Before empty relationships was included. With this PR they are not. I can't find if the spec says one of these two is the correct behavior. 

@joukevandermaas can you clarify "Is there some way to dis-allow the client from including resources? The endpoint still has to support it for it to work."?

Tests/JsonApiSerializerTests.cs
bjornharrtell: Sorry, I don't quite understand what your meaning is here.

joukevandermaas: I think there needs to be a test that verifies that the `included` hash in the JSON API response has only one version of each resource, even if multiple resources refer to it.

Example:

1. Two people both have a job at the same company
2. Request all people and ask for the job to be included
3. Verify that the `included` hash in the JSON API document only has the company once.

joukevandermaas:
@bjornharrtell

1. I think that's fine in theory, but it's a breaking change. I'd rather keep the current behavior and allow users of Saule (API writers) to opt into the new behavior, e.g. with an `[AllowClientInclude]` attribute on the action method, or a parameter to the `BelongsTo` and `HasMany` functions in `ApiResource`.

This is also what I meant by *"is there some way to dis-allow the client from including resources? The endpoint still has to support it for it to work."* The API writer should be able to control whether the client can or cannot include a given relationship.

2. IIRC empty relationships were converted into a link (they were serialized with a `links` hash that allowed clients to fetch the data async).

```json
{
  "attributes": { ... },
  "relationships": {
    "job": {
      "links": {
        "related": "http://.../job",
        "self": "http://.../relationships/job"
      }
    }
  }
}
```

I think that should still happen if the client did not ask to include the resource or if the resource property does not exist on the model.

bjornharrtell:
Thanks for the explanations, I think it makes sense.

yohanmishkin:
1. We had some heavy objects with many nested relationships so that's why I made inclusions opt-in by default, but it makes sense to me to avoid introducing breaking changes.

2. To be honest, I'm not positive why I removed empty relationships, but I think it may have been ember-cli-mirage inspired.

bjornharrtell:
Ready for review once more @joukevandermaas. Things done:

* Should now be without any breaking change for API consumers
* Default is to include related data but can be opt-out via action attribute (this concerns only the case where the include param is not used)
* Added test for included resource referenced in multiple resources

.gitattributes
joukevandermaas: Just keep it removed in this PR and I'll add it back later so it doesn't clutter up the diff.

Saule/Http/NoDefaultIncludedAttribute.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Web.Http.Controllers;
+using System.Web.Http.Filters;
+
+namespace Saule.Http
+{
+    /// <summary>
+    /// Indicates that the action should not include related data by default.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method)]
+    public sealed class NoDefaultIncludedAttribute : ActionFilterAttribute
```
joukevandermaas: Perhaps rename this `AllowsIncludesAttribute` and move that functionality from `AllowsQueryAttribute` into here? Let me know what you think. The documentation can still explain that this also disables the includes by default.

Saule/JsonApiSerializer.cs
```diff
@@ -72,7 +72,9 @@ public PreprocessResult PreprocessContent(object @object, ApiResource resource,
                     type: resource,
                     baseUrl: requestUri,
                     urlBuilder: UrlPathBuilder,
-                    paginationContext: QueryContext?.Pagination);
+                    paginationContext: QueryContext?.Pagination,
+                    includingContext: QueryContext?.Including,
+                    includedDefault: QueryContext != null ? QueryContext.IncludedDefault : true);
```
joukevandermaas: I understand why this is a parameter (query context may always be `null`), but I'd prefer if `null` was just interpreted as `true` in the code, with a comment. There's probably some edge case I can't think of where that doesn't work, but if it does then I'd prefer not adding an extra parameter.

Saule/StringExtensions.cs
```diff
@@ -23,7 +23,7 @@ public static string EnsureStartsWith(this string source, string start)
 
         public static string TrimJoin(this char separator, params string[] parts)
         {
-            return string.Join(separator.ToString(), parts.Select(p => p.Trim(separator)));
+            return string.Join(separator.ToString(), parts.Select(p => p?.Trim(separator)));
```
joukevandermaas: In what situation does parts contain `null`? Would we not just want to filter those out instead?

bjornharrtell: Not sure, it was from https://github.com/joukevandermaas/saule/pull/136/commits/368ea27c0d21d5c4bd9050b52ac3aef460afa8ec by @yohanmishkin.

Saule/Http/NoDefaultIncludedAttribute.cs
bjornharrtell: I'm not entirely sure what you mean. You mean to say that allowing querying should implicitly allow the include param (or the other way round) and at the same time disable the default include? I think something like that might make sense.

Saule/StringExtensions.cs
yohanmishkin: Totally can't remember! I'm so helpful.

Saule/Http/NoDefaultIncludedAttribute.cs
joukevandermaas: Yeah, after thinking about this for a bit longer I think your initial idea was right. I think something like this should work:

| |`AllowsQuery` off |`AllowsQuery` on|
|---|---|---|
|**`NoDefaultIncluded` off** | Don't allow `include` query; always include all available relationships | When `include` provided, include only those relationships, otherwise everything |
|**`NoDefaultIncluded` on** | Don't allow `include` query; never include any available relationships | When `include` provided, include only those relationships, otherwise nothing |

The lower left here is a bit strange, and could be changed if that makes sense. I'm not a huge fan of the name `NoDefaultIncluded`, because all other attributes have an active verb (e.g. `ReturnsResource`, `AllowsQuery`). Unless you can think of something better I think it should be fine though :smile:

Saule/JsonApiSerializer.cs
bjornharrtell: If I understand you correctly (I don't quite understand the "may always be null" part..) that would require the parameter/member includedDefault in ResourceSerializer to be a nullable boolean, isn't it better to coalesce the QueryContext.IncludedDefault before that as done here?

Saule/StringExtensions.cs
bjornharrtell: Well, all tests passes without this change so I'll revert it and if there is a reason we will find it eventually. :)

Saule/Http/NoDefaultIncludedAttribute.cs
bjornharrtell: I suggest the name `DisableDefaultIncluded`.

Saule/JsonApiSerializer.cs
bjornharrtell: Looking at this again there might be some confusion here. I now see that the `QueryContext` was intended for only the querystring parameters for the request. I thought about the word Query in a broader context i.e something like Request.

I added `IncludedDefault` without considering the above. Perhaps it would be better to separate the `IncludedDefault` state into another property of the `Request.Properties` of the `HttpActionContext` but then again it might not be worth that extra separation. What is your opinion @joukevandermaas?

joukevandermaas: Really all I meant was that everything related to including resources should live on the including context; IMO there is no need for a separate including-related flag on the query context itself. That way, you always know you have all relevant information when you have the including context.

bjornharrtell: Ah yes, that more simple. I'll have a go at it later and report back.

bjornharrtell:
I've refactored to keep the state of `DisableDefaultIncluded` on `IncludedContext` instead of `QueryContext`, and assuming `false` if `IncludedContext` is `null`. Ready for review yet again @joukevandermaas.

joukevandermaas:
@bjornharrtell Thanks so much for seeing this through! I'm going to merge this and publish a new beta release with the changes.

@yohanmishkin Thanks for your contributions to this!

Issue #140, opened 2017-02-23
Autogenerated Swashbuckle swagger documentation is incorrect


bjornharrtell:
As Swashbuckle will try to guess the schema for the plain entity types it will not expose a swagger documentation for the actual JSON API as exposed by Saule.

It would be nice to somehow make Swashbuckle understand that what is really input/output in the REST API is ApiResources, without customizing the Swashbuckle Config.

bjornharrtell:
While it would be good to auto wire this somehow, at least it seems possible to use SwaggerResponseAttribute and SwaggerSchemaFilterAttribute to correctly instruct Swashbuckle.

bjornharrtell:
I've had success implementing special purpose `IDocumentFilter` and `IOperationFilter` which could be packaged into a `ISwaggerProvider` for a more or less generic solution to auto wire support for JSON API with Swashbuckle. The latter would require further non-trivial work.

A caveat is that it's not possible to ref/extend to the JSON API JSON Schema as Swagger 2.0 does not support enough of JSON Schema. This seems to be changing in Swagger 3.0 but it's far off to be generally available. I still think Swagger 2.0 can be useful as documentation for JSON API, you just have to leave out some details which can be referred to the actual JSON API spec at relevant places.

joukevandermaas:
I think this was solved with #142.

PR #142, opened 2017-03-21
Expose properties of resource

bjornharrtell:
Needed for external inspection to for example generate schema.

Can solve part of #140.

### DIFF
--- a/Saule/ApiResource.cs
+++ b/Saule/ApiResource.cs
@@ -43,11 +43,11 @@ namespace Saule
         /// </summary>
         public string ResourceType { get; private set; }
 
-        internal IEnumerable<ResourceAttribute> Attributes => _attributes;
+        public IEnumerable<ResourceAttribute> Attributes => _attributes;
 
-        internal IEnumerable<ResourceRelationship> Relationships => _relationships;
+        public IEnumerable<ResourceRelationship> Relationships => _relationships;
 
-        internal string IdProperty { get; private set; }
+        public string IdProperty { get; private set; }
 
         /// <summary>
         /// Customize the type name of this resource. The default value
--- a/Saule/ApiResource.cs
+++ b/Saule/ApiResource.cs
@@ -44,17 +44,17 @@ namespace Saule
         public string ResourceType { get; private set; }
 
         /// <summary>
-        /// The defined attributes of this resource.
+        /// Gets the defined attributes of this resource.
         /// </summary>
         public IEnumerable<ResourceAttribute> Attributes => _attributes;
 
         /// <summary>
-        /// The defined relationships of this resource.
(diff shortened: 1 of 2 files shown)


joukevandermaas:
I'm not entirely sure why this would help with #140, since I've never used Swashbuckle before. I'm generally pretty hesitant to making things public, but I'm assuming you have a specific use in mind for this API. 

Can you maybe explain with a code snippet or whatever?

bjornharrtell:
I don't have a good code snippet to show but I hope I can explain the purpose better without requiring previous knowledge. So Swashbuckle works by introspecting the Web API controllers, to generate a swagger document which essentially is a schema/documentation of the REST API. The built in introspection in Swashbuckle will only see the type as returned by an `ApiController`, even if it is transformed by Saule the corresponding `ApiResource`, because it does not understand or consider the `ReturnsResourceAttribute` from Saule.

However, I can provide code to customize the introspection. In that code I can find the actual `ReturnsResourceAttribute` instance for the controller and then proceed to customize the schema to represent a JSON API resource instead of the plain JSON representation of a model. To be able to implement the actual customized schema I need to be able to look at the defined `Attributes`,  `ResourcesRelationships` and `IdProperty` on the `ApiResource` which is why I want them exposed.

For now I only have a hacked solution for the actual customization logic but I think that I will eventually be able to provide an extension to Swashbuckle that can easily be used in any Saule project. This would be similar to https://github.com/rbeauchamp/Swashbuckle.OData which provides something similar but for OData.

joukevandermaas:
Ok, that sounds good. I'll merge this if you can rebase on `master`. Thanks for the explanation!

bjornharrtell:
Rebased. I've also added a property OriginalName to ResourceAttribute as that too is useful the context of introspection for Swashbuckle, I hope that's acceptable.

joukevandermaas:
Since people can pass any string to this constructor (through `Attribute` in `ApiResource`), I wonder if we should still normalize it, but to PascalCase (conventional for C# properties). That way Swashbuckle can get the property name (assuming that is the intention) and people can still specify the attributes in JSON API form if they want.

Of course if we do that, it should be called something like `PropertyName`, not `OriginalName`. Relationships (`ResourceRelationship`) should probably also get this property, just to be consistent.

Does this make sense to you? Am I correctly understanding the purpose of this new property?

bjornharrtell:
Yes I think it would make sense, but it would also break existing public API?

bjornharrtell:
Ah no, I misread you and now it makes even more sense. I'll try to make the changes as suggested.

bjornharrtell:
@joukevandermaas I hope this is good to go now.

joukevandermaas:
Yup! thanks for seeing this through!

PR #147, opened 2017-04-21
Prevent a resource being included in both the data and included sections

laurence79:
This addresses a problem I came across when Saule is serialising an array of resources. If any of the resources in the array have relationships with each other, then it will also add the related resources to the included section. 

The JSON:API spec asks that resources in compound documents should be unique across the data and included sections. http://jsonapi.org/format/#document-compound-documents

This PR trims resources from the included section where they are present in the data section once resource serialisation is complete.


joukevandermaas:
Awesome, thanks so much for this!

Issue #141, opened 2017-03-21
Output of id is of type string in Json-Api, but is outputted as original type.


Guidhouse:
```c#    
public class User
    {
        public long UserId { get; set; }//For reasons of the DBA's we cannot change this easily
        public string Description { get; set; }
    }
```
Using ApiResource with 'WithId("UserId")' and the JsonApiSerializer<UserResource> captures the Id for the schema.@default ( in Swashbuckler), but the value is outputted without the quotation marks mandated by Json-Api, as seen here: http://jsonapi.org/format/#document-resource-objects.

what we get is the following json:

```json
default": {
    "data": {
        "type": "user",
        "id": 1,
        "attributes": {
            "description": "Test"
        }, ...
```
I hope my explanation is clear enough, and if I get the time I will try and fix it myself and make a pull-request.

joukevandermaas:
Good catch! A PR would be much appreciated.

joukevandermaas:
Fixed in #149.

Issue #99, opened 2016-05-05
Object(s) are serialized to JSON before applying the resource attribute filters. (Huge problem unless a resource is detached)


NullVoxPopuli:
So, I have an EF object.

My issue is that in `ResourceSerializer`, this line exists:

``` csharp
var objectJson = JToken.FromObject(_value, serializer);
```

This traverses ALL the attributes and relationships on the EF object.

What should happen is that based on the defined attributes and relationships in the Resource, is that those should iterate over the object reflectively pulling the attributes and relationships off the object. So that way, we don't unnecessarily serialize a giant tree of things (especially if lazy loading is enabled (I know people frown on lazy loading, that's not a discussion for here))

/cc @wozbo

joukevandermaas:
Yeah I agree, this is not the desired behavior. Sadly, it would be non-trivial to change it. Maybe you can reduce the size of your EF object until I get a chance to pick this up?

NullVoxPopuli:
I'm working on a solution, too. I have pushed the branch, but am still
working on it

joukevandermaas:
Awesome, thanks!

rhyek:
Hello, gents. I also need this behavior for a new project I'm currently working on. Any idea when this might be implemented? Thanks!

joukevandermaas:
@rhyek: @NullVoxPopuli has been doing some work in #103. I'm not sure of the status of this, but the build is currently failing so I cannot merge it.

joukevandermaas:
Fixed in #149.

PR #149, opened 2017-05-10
Serialise complex objects correctly

laurence79:
This is essentially a reimplementation of the ResourceSerializer that copes with a couple of edge cases that have tripped me up in my own use:
- Circular references
- Inclusion of deeply nested objects

Take a look at ResourceGraphSerializerTests to see tests that pass with the new serialiser. The new serialiser also passes all the existing tests (with a couple of exceptions where I have overridden the existing test with comments as to why).

The new serialiser approach differs from the original by first building a graph of objects to serialise using the IncludingContext, and then generates the response JSON from that.

This is anecdotal but I've also seen modest performance improvements from this approach.

By default the original serialiser is still used unless a new configuration option is set
`
config.ConfigureJsonApi(new JsonApiConfiguration
{
    UseGraphSerializer = true
});
`

I appreciate this is a large change, but hopefully it will be a benefit to the project 😊


joukevandermaas:
This sounds great, I'll try to make some time to review this soon. Thanks so much!

laurence79:
Thanks @joukevandermaas

laurence79:
Fixes #141

bjornharrtell:
Seems like nice work @laurence79, looking forward to testing it. I wonder if it also resolves https://github.com/joukevandermaas/saule/issues/99?

laurence79:
Thanks @bjornharrtell, yes it seems like it might because only the attributes and relationships declared in the resource are read from the target objects.

Saule/Http/JsonApiConfiguration.cs
```diff
@@ -22,6 +22,11 @@ public class JsonApiConfiguration
         /// <summary>
         /// Gets the expressions that are used to evaluate filter queries on a per-type basis.
         /// </summary>
-        public QueryFilterExpressionCollection QueryFilterExpressions { get; } = new QueryFilterExpressionCollection();
+        public QueryFilterExpressionCollection QueryFilterExpressions { get; } = new QueryFilterExpressionCollection();
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the new experimental graph serializer is used.
+        /// </summary>
+        public bool UseGraphSerializer { get; set; } = false;
```
joukevandermaas: I wonder if we actually need this; I would prefer we just swap out the serializer (provided that all the tests pass with the new one).

I don't think anyone would want to use the old serializer with this new one in place anyway.

Saule/Http/AllowsIncludeAttribute.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// No related resources should be included.
+        /// </summary>
+        None
+    }
+
+    /// <summary>
+    /// Indicates that the target action supports the use of the include query param.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method)]
+    public sealed class AllowsIncludeAttribute : ActionFilterAttribute
```
joukevandermaas: Moving this out of `AllowsQueryAttribute` seems unrelated to serialization. What was the reason for this change?

Since I already released 1.6, this would be a breaking change. I don't really mind (we can figure out a migration scenario or just release 2.0), but I'm not sure I understand the purpose of this change from the diff.

Saule/ObjectExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+            }
+
+            var propertyInfo = obj.GetType().GetProperty(propertyName);
+            if (propertyInfo != null)
+            {
+                return propertyInfo.GetValue(obj);
+            }
+
+            var fieldInfo = obj.GetType().GetField(propertyName);
+            if (fieldInfo != null)
```
joukevandermaas: I'm not sure we should be reading fields. Unless this is actually used somewhere, I'd prefer we remove it.

Saule/Serialization/ResourceGraphNodeKey.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Saule.Serialization
+{
+    internal class ResourceGraphNodeKey : IEquatable<ResourceGraphNodeKey>
```
joukevandermaas: I'm thinking this could be a `struct`?

Saule/Serialization/ResourceGraph.cs
```diff
@@ -0,0 +1,144 @@
...
+            ResourceGraphPathSet includePaths,
+            int depth)
+        {
+            if (obj == null)
+            {
+                // end of the line - obj will be a leaf node in our graph
+                return;
+            }
+
+            if (obj is IEnumerable)
```
joukevandermaas: How does this work with `IQueryable`? Does it evaluate the query to build the graph? If so, wouldn't that mean multiple queries to fetch the relationships? Maybe we need some tests for this stuff, since it seems pretty easy to break with a simple change.

Tests/Serialization/ResourceGraphSerializerTests.cs
```diff
@@ -0,0 +1,214 @@
...
+            var target = GetSerializer(person, new PersonWithDefaultIdResource(),
+                GetUri(id: "123"), DefaultPathBuilder, null, null);
+            var result = target.Serialize();
+            _output.WriteLine(result.ToString());
+
+            var relationships = result["data"]["relationships"];
+            var job = relationships["job"];
+            var friends = relationships["friends"];
+
+            Assert.Equal(job["data"].Type.ToString(), "Null");
```
joukevandermaas: I think you can just assert on the enum value, i.e.: 

```csharp
Assert.Equal(job["data"].Type, JTokenType.Null);
```

Tests/Serialization/ResourceGraphSerializerTests.cs
```diff
@@ -0,0 +1,214 @@
...
+            Assert.NotNull(friends);
+        }
+
+        public override void HandlesNullValues()
+        {
+            var person = new Person(id: "45");
+            var target = GetSerializer(person, DefaultResource,
+                GetUri(id: "45"), DefaultPathBuilder, null, new IncludingContext { DisableDefaultIncluded = true });
+            /**
+             *  Changed from a null IncludingContext in the original version
```
joukevandermaas: Does this mean the `IncludingContext` is never `null` anymore (seems fine, just wondering)?

Tests/Serialization/ResourceSerializerTests.cs
```diff
@@ -9,25 +9,39 @@
...
 using Saule.Queries.Including;
+using Saule.Queries.Pagination;
 
 namespace Tests.Serialization
 {
+    [Trait("Serializer", "ResourceSerializer")]
     public class ResourceSerializerTests
     {
-        private readonly ITestOutputHelper _output;
+        internal readonly ITestOutputHelper _output;
```
joukevandermaas: Maybe you mean `protected`? In any case, if we just remove the old serializer, this should not be necessary anymore.

Saule/Http/JsonApiConfiguration.cs
laurence79: Sure, I've moved the new to be in place of the old and removed the config option.

Saule/Serialization/ResourceGraphNodeKey.cs
laurence79: Agreed, change made.

Tests/Serialization/ResourceGraphSerializerTests.cs
laurence79: Of course, silly me 🙄

Tests/Serialization/ResourceSerializerTests.cs
laurence79: Old serialiser replaced.

laurence79:
AppVeyor had a timeout on package restore. I couldn't see a way to easily retry the build, so I bumped the Newtonsoft.Json version to 1.0.2, hope thats ok 😊

Saule/ObjectExtensions.cs
laurence79: I didn't originally have this, but some tests failed e.g. "Handles recursive properties on resource objects", and I figured the old serialiser would serialise fields.

Saule/Serialization/ResourceGraph.cs
laurence79: Good point, I'll write some tests to see how this behaves.

Tests/Serialization/ResourceGraphSerializerTests.cs
laurence79: No, I think the purpose of this test was to prove that if DisableDefaultIncluded was true then no relationships were returned.

It's entirely possible that I've misread this of course 😆

Saule/Http/AllowsIncludeAttribute.cs
laurence79: Good spot. This is not needed, I've reverted it to how it was before.

Saule/Serialization/ResourceGraph.cs
laurence79: Ok perhaps I'm not completely clear on this. 

The expression tree is already evaluated and a response generated before the graph starts reflecting upon it. Is this what you meant @joukevandermaas?

joukevandermaas: I was half-referring to #99. If I have an object graph like this, where every property is lazy (say, using an ORM):

```
- Person
  - Job
    - LiterallyEveryEmailEverSent     // huge!
```

but no Saule resource ever refers to `LiterallyEveryEmailEverSent`:

```csharp
class PersonResource : ApiResource
{
  public PersonResource() {
    BelongsTo<CompanyResource>(nameof(Person.Job));
  }
}

class CompanyResource : ApiResource
{
  public CompanyResource() {
  }
}
```

The current serializer will materialize `LiterallyEveryEmailEverSent`, which is obviously wasteful because it won't end up in the final JSON output. Does the new serializer still have this effect, or is it circumvented somehow? It's not immediately obvious to me from the code (these things can be tricky).

Since the current serializer *does* materialize everything, I think it's OK if the new one does too; we can always improve on it later. That said, if it already *doesn't* do it, that would be better :smile: In that case, I think tests are necessary, because it seems pretty easy to mess up somewhere and evaluate everything when someone changes something later.

laurence79: Ok understood. The new serialiser only reads object properties that are declared as attributes or relationships in the resource. So in this case, `LiterallyEveryEmailEverSent` will not be read. I've tidied up the code slightly to make this intent clearer.

I've added a test case to cover this, and also added functionality to serialise dictionaries and dynamics. The old serialiser would have handled this, but originally there were no tests for this specifically.

laurence79:
Ok @joukevandermaas, I think this is ready to be reviewed again 😊

joukevandermaas:
@laurence79 Thanks a lot for this! I published a new pre-release with these changes in it.

Issue #146, opened 2017-04-21
Should not require Content-Type application/vnd.api+json for DELETE


bjornharrtell:
* http://jsonapi.org/format/#crud-deleting

adamalesandro:
@bjornharrtell I have a related type of issue. In our app we have one route that we would like to serve without requiring `application/vnd.api+json` Accept header. For my solution, I was thinking adding additional logic to ReturnsResourceAttribute to effectively turn off that requirement.

ReturnsResourceAttribute constructor:
```cs
public ReturnsResourceAttribute(Type resourceType, bool disableJsonApiFilter = false)
```

Usage:
```cs
[Route("api/attachments/{id}/download")]
[ReturnsResource(typeof(AttachmentResource), disableJsonApiFilter: true)]
public async Task<HttpResponseMessage> DownloadAttachment(int id)
{
...// etc.
```

While I don't expect that you would decorate all your DELETE actions this way, perhaps there is some common ground here. @joukevandermaas, curious if you've thought about this at all.

bjornharrtell:
It's similar to https://github.com/joukevandermaas/saule/issues/100. I hope to be able to contribute a PR to fix it eventually.

joukevandermaas:
@adamalesandro In my understanding, the spec says it's required if the client is talking to a JSON API server. There should be a way to 'turn off' Saule for specific endpoints, but then it should also turn off serialization IMO. I'm curious about your use-case, though.

@bjornharrtell I think Saule should not require the `content-type` header for *any* request without content (including `DELETE` of course), I think. At least, this makes sense to me :smile:

bjornharrtell:
Agreed @joukevandermaas. I think @adamalesandro is also talking about requests without content.

adamalesandro:
Yes, correct. In our case we want to return a file (as download) from an href link in an ember app.

adamalesandro:
I should also add that I'm new to ember/JSON-API so if there's an already accepted way of serving files, I'd go that way. I have not come across anything viable from my research yet.

bjornharrtell:
I was refering to posting (or patching) content to the server. Delivering content from the server anything else than `application/vnd.api+json` isn't JSON API, so I'd say Saule should not be involved - just make an ordinary Web API 2 controller.

bjornharrtell:
@adamalesandro I see you are refering to both `Accept` and `Content-Type` headers which is confusing.

[...]

Issue #100, opened 2016-05-05
406 when making a request without the json-api media type


jonnii:
Hello! It's really frustrating to test any API using saule in a browser as it will 406 unless the content type matches exactly the json-api media type. I'm unsure as to why you made the decision as my reading of the Content Negotiation section of the spec (http://jsonapi.org/format/#content-negotiation) doesn't seem to indicate that this is required?

It _does_ say that if the content is supplied how it should be handled, but doesn't seem to go further than that.

NullVoxPopuli:
it says this though:

```
Clients MUST send all JSON API data in request documents with the header Content-Type: application/vnd.api+json without any media type parameters.
```

is that what you are not wanting?

jonnii:
The way I read that is is that if the content-type application/vnd.api+json is specified then it must not contain any media type parameters. Perhaps I'm wrong... 

Either way not being able to use a browser to view the API is kind of awkward. The fact that saule also overwrites all the existing formatters and that `PreprocessingDelegatingHandler` is internal means it's not easy to setup the formatting in such a way to fit into an existing application which might also serve json and/or other content types.

jonnii:
And, one more thing. If you try to use it with something like swagger (to allow you to view and test your API) then it does weird things as the other formatters have been removed. It would be great if we could use saule and swagger in the same project.

NullVoxPopuli:
Yeah, there is def room for flexability with this. I'd like something similar to active_model_serializers.

have you used postman? that allows setting the content-type

joukevandermaas:
I definitely agree you should be able to serve other content types from a project that uses Saule. Maybe an overload to `ConfigureJsonApi` that inserts the json api formatter at the start of the collection instead?

jonnii:
swagger lets you set (and even detects) the media-type compatible with your API, it's pretty advanced. i can see the similarity to AMS, it's definitely a good approach! Right now we are returning anonymous types from our controllers, but there's a lot of boilerplate =(

joukevandermaas:
#101 should make configuration more flexible when it's merged. 

I'm not 100% sure on how to read the spec regarding the original issue, but I think we shouldn't change behavior until we get clarification from the authors. Feel free to open an issue over at [the spec's repo](https://github.com/json-api/json-api).

jonnii:
There's already an issue discussing this:

https://github.com/json-api/json-api/issues/968

joukevandermaas:
[...]

PR #151, opened 2017-05-12
Do not require Accept header

bjornharrtell:
My take on https://github.com/joukevandermaas/saule/issues/100 and https://github.com/joukevandermaas/saule/issues/146.

### DIFF
--- a/Saule/Http/ReturnsResourceAttribute.cs
+++ b/Saule/Http/ReturnsResourceAttribute.cs
@@ -40,7 +40,7 @@ namespace Saule.Http
         {
             var accept = actionContext.Request.Headers.Accept
                 .Where(a => a.MediaType == Constants.MediaType);
-            if (accept.All(a => a.Parameters.Any()))
+            if (accept.Count() > 0 && accept.All(a => a.Parameters.Any()))
             {
                 // no json api media type without parameters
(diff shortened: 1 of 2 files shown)


joukevandermaas:
Thanks for this!

I understood the requirements from #100 and #146 as follows:

1. If the client does not send a request body, it does not need to send a `Content-Type` header
2. The client does not need to send an accept header
3. If the client *does* send an accept header, it must be `application/vnd.api+json` without any media type parameters

I don't really see these reflected in this PR though. My understanding of the code is these rules:

4. The client does not need to send an accept header for `GET` and `DELETE` requests
5. Other requests must have an accept header, and it must be `application/vnd.api+json` without any media type parameters
6. Nothing about `Content-Type` in the diff.

4 and 5 don't make much sense to me, because I thought the `Accept` header was about the `content-type` of the *response*, not the request. It makes sense to validate it always (if it exists).

If I misunderstood something, please let me know!

bjornharrtell:
You are absolutely correct, for some reason I mixed up these headers logically when taking a stab at this.

I suggest a more simple change to simply let requests through that have no `application/vnd.api+json` `Accept` header at all. For `Content-Type` I'm not sure there is an issue at all.

bjornharrtell:
Reworked as suggested. I hope it makes sense this time.

joukevandermaas:
Looks good! thanks!

PR #152, opened 2017-05-16
Remove links property assignment that unused and overwritten

bjornharrtell:


joukevandermaas:
@bjornharrtell I think this is obsolete with #149?

bjornharrtell:
The old `ResourceSerializer` is still there so might still be relevant, if you are not intending to remove it in short term?

joukevandermaas:
It is not there anymore, we removed it. The new one is so much better! :smile:

bjornharrtell:
Great to hear it and I look forward to try it out! :) Closing.

Issue #121, opened 2016-06-27
Support optional fields in resource object


bxh:
Per JSON API [spec](https://github.com/joukevandermaas/saule/blob/master/Saule/Serialization/ResourceSerializer.cs#L179), `attributes`, `relationships` and `links` are optional. However, in Saule they are always provided. This could cause unexpected behaviors for JSON API clients such as Ember Data. And although `link` is a neat feature, it takes up additional space. I'm wondering if the community here is willing to make those fields optional like `included`?

joukevandermaas:
I'm very much in favor of leaving out optional keys when they are empty. Sounds like a good feature to me.

PR #153, opened 2017-05-17
Optional links support

### LINKED ISSUE #150: Make links optional
According to the specification links are optional, see:

* http://jsonapi.org/format/#document-top-level
* http://jsonapi.org/format/#document-resource-objects

I've encountered situations where I would like them not included. Saule autogenerates links and has AFAIK no option to disable them all or selectively. Would it make sense to introduce such a feature?

I have tried overriding as described by https://github.com/joukevandermaas/saule/wiki/Generating-links but even if returning null there will be an empty links objekt. Also, that way seem a bit overkill and I'm thinking an attribute could perhaps be created for this purpose instead.

bjornharrtell:
Aims to resolve #150.

Not yet finished though but wanted to ask about the approach. I couldn't see it possible to handle this in the `UrlPathBuilder` but had to involve the `ResourceSerializer`. Also, I'm not sure about what a "canonical" url is and so have problems understanding the purpose of Related vs RelatedCanonical.

### DIFF
--- a/Saule/ApiResource.cs
+++ b/Saule/ApiResource.cs
@@ -6,6 +6,38 @@ using Humanizer;
 namespace Saule
 {
     /// <summary>
+    /// Describes one or more link types to be generated.
+    /// </summary>
+    [Flags]
+    public enum LinkType
+    {
+        /// <summary>
+        /// No links
+        /// </summary>
+        None = 0,
+
+        /// <summary>
+        /// Only self links
+        /// </summary>
+        Self = 1,
+
+        /// <summary>
+        /// Only related links
+        /// </summary>
+        Related = 2,
+
+        /// <summary>
+        /// Only ? links
+        /// </summary>
+        RelatedCanonical = 4,
+
+        /// <summary>
+        /// Generate all possible links
+        /// </summary>
+        All = ~None
+    }
+
+    /// <summary>
     /// Represents a resource that can be consumed by clients.
     /// </summary>
     public abstract class ApiResource
@@ -59,6 +91,11 @@ namespace Saule
         public string IdProperty { get; private set; }
 
         /// <summary>
+        /// Gets the defined <see cref="LinkType"/> to be generated for this resource.
+        /// </summary>
+        public LinkType LinkType { get; private set; } = LinkType.All;
+
+        /// <summary>
         /// Customize the type name of this resource. The default value
         /// is the name of the class (without 'Resource', if it exists).
         /// </summary>
@@ -94,6 +131,19 @@ namespace Saule
             IdProperty = name.ToPascalCase();
 
             return IdProperty;
+        }
+
+        /// <summary>
+        /// Customize the LinkType property of this resource. The default value
+        /// is <see cref="LinkType.All"/>.
+        /// </summary>
+        /// <param name="withLinks">The desired <see cref="LinkType"/> to generate for this resource.</param>
+        /// <returns>Value that was set.</returns>
+        protected LinkType WithLinks(LinkType withLinks)
+        {
+            LinkType = withLinks;
+
+            return LinkType;
         }
 
         /// <summary>
@@ -135,10 +185,25 @@ namespace Saule
         protected ResourceRelationship BelongsTo<T>(string name, string path)
                     where T : ApiResource, new()
         {
+            return BelongsTo<T>(name, path, LinkType.All);
+        }
+
+        /// <summary>
+        /// Specify a to-one relationship of this resource.
+        /// </summary>
+        /// <param name="name">The name of the relationship.</param>
+        /// <param name="path">The url pathspec of this relationship (default
+        /// is the name)</param>
+        /// <param name="withLinks">The defined <see cref="LinkType"/> to be generated for this relationship.</param>
+        /// <typeparam name="T">The api resource type of the relationship.</typeparam>
+        /// <returns>The <see cref="ResourceRelationship"/>.</returns>
+        protected ResourceRelationship BelongsTo<T>(string name, string path, LinkType withLinks)
+                    where T : ApiResource, new()
+        {
             VerifyPropertyName(name);
 
             var resource = GetUniqueResource<T>();
-            var result = new ResourceRelationship<T>(name, path, RelationshipKind.BelongsTo, resource);
+            var result = new ResourceRelationship<T>(name, path, RelationshipKind.BelongsTo, resource, withLinks);
 
             _relationships.Add(result);
 
@@ -167,10 +232,24 @@ namespace Saule
         protected ResourceRelationship HasMany<T>(string name, string path)
                     where T : ApiResource, new()
         {
+            return HasMany<T>(name, name, LinkType.All);
+        }
+
+        /// <summary>
+        /// Specify a to-many relationship of this resource.
+        /// </summary>
+        /// <param name="name">The name of the relationship.</param>
+        /// <param name="path">The url pathspec of this relationship (default is the name).</param>
+        /// <param name="withLinks">The defined <see cref="LinkType"/> to be generated for this relationship.</param>
+        /// <typeparam name="T">The api resource type of the relationship.</typeparam>
+        /// <returns>The <see cref="ResourceRelationship"/>.</returns>
+        protected ResourceRelationship HasMany<T>(string name, string path, LinkType withLinks)
+                    where T : ApiResource, new()
+        {
             VerifyPropertyName(name);
 
             var resource = GetUniqueResource<T>();
-            var result = new ResourceRelationship<T>(name, path, RelationshipKind.HasMany, resource);
+            var result = new ResourceRelationship<T>(name, path, RelationshipKind.HasMany, resource, withLinks);
 
             _relationships.Add(result);
 
--- a/Saule/ApiResource.cs
+++ b/Saule/ApiResource.cs
@@ -27,11 +27,6 @@ namespace Saule
         Related = 2,
 
         /// <summary>
-        /// Only ? links
-        /// </summary>
-        RelatedCanonical = 4,
-
-        /// <summary>
         /// Generate all possible links
         /// </summary>
         All = ~None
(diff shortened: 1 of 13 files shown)


Saule/ApiResource.cs
```diff
@@ -58,6 +68,8 @@ protected ApiResource()
         /// </summary>
         public string IdProperty { get; private set; }
 
+        public LinkTypes WithLinks { get; set; } = LinkTypes.All;
```
joukevandermaas: I don't think `ApiResource` should have any mutable properties.

Saule/Serialization/ResourceSerializer.cs
```diff
@@ -96,10 +97,12 @@ private ResourceGraphPathSet IncludedGraphPathsFromContext(IncludingContext cont
 
         private JToken CreateTopLevelLinks(int count)
         {
-            var result = new JObject
+            var result = new JObject();
+
+            if (_resource.WithLinks.HasFlag(LinkTypes.Self & LinkTypes.All))
```
joukevandermaas: I think you can just do `HasFlag(LinkTypes.Self)`, since `All` includes that. Now this is only added when the value is `LinkTypes.All`.

Saule/Serialization/ResourceSerializer.cs
```diff
@@ -235,24 +246,39 @@ private JObject SerializeRelationships(ResourceGraphNode node)
...
                 }
 
                 var links = new JObject();
-                AddUrl(links, "self", _urlBuilder.BuildRelationshipPath(node.Resource, node.Key.Id.ToString(), kv.Value.Relationship));
-                AddUrl(links, "related", _urlBuilder.BuildRelationshipPath(node.Resource, node.Key.Id.ToString(), kv.Value.Relationship, relationshipId));
+                var self = _urlBuilder.BuildRelationshipPath(node.Resource, node.Key.Id.ToString(), relationship);
+                var related = _urlBuilder.BuildRelationshipPath(node.Resource, node.Key.Id.ToString(), relationship, relationshipId);
 
-                item["links"] = links;
+                if (!string.IsNullOrEmpty(self) && relationship.WithLinks.HasFlag(LinkTypes.Related & LinkTypes.All))
```
joukevandermaas: Same here, you don't need to do the `&`.

Saule/Serialization/ResourceSerializer.cs
```diff
@@ -235,24 +246,39 @@ private JObject SerializeRelationships(ResourceGraphNode node)
...
+                {
+                    AddUrl(links, "self", self);
+                }
+
+                if (!string.IsNullOrEmpty(related) && relationship.WithLinks.HasFlag(LinkTypes.Related & LinkTypes.All))
```
joukevandermaas: And here.

bjornharrtell:
Thanks @joukevandermaas, good to know I'm hopefully on the right path. Will ping you when I have had the time to make this PR complete and properly reviewable and will take your early comments into consideration.

bjornharrtell:
Improved now but still tests still too few and should try it out on a real API implementation too.

bjornharrtell:
Additional tests added and seems to do what I expect in a real project. Ready for review now @joukevandermaas.

But... I did notice a possible regression with the new serializer. Using 1.7.0.390 I'm missing the `data` property for relationships (only tested BelongsTo as of yet), downgrading to stable 1.6 resolves it.

laurence79:
@bjornharrtell damn... could you give me an example or even better, a test, and I'll investigate the regression?

bjornharrtell:
@laurence79 will try to produce a unit test to demonstrate the issue.

bjornharrtell:
See testcase in https://github.com/joukevandermaas/saule/pull/155.

Saule/Serialization/ResourceSerializer.cs
```diff
@@ -235,24 +248,39 @@ private JObject SerializeRelationships(ResourceGraphNode node)
...
                 }
 
                 var links = new JObject();
-                AddUrl(links, "self", _urlBuilder.BuildRelationshipPath(node.Resource, node.Key.Id.ToString(), kv.Value.Relationship));
-                AddUrl(links, "related", _urlBuilder.BuildRelationshipPath(node.Resource, node.Key.Id.ToString(), kv.Value.Relationship, relationshipId));
+                var self = _urlBuilder.BuildRelationshipPath(node.Resource, node.Key.Id.ToString(), relationship);
+                var related = _urlBuilder.BuildRelationshipPath(node.Resource, node.Key.Id.ToString(), relationship, relationshipId);
 
-                item["links"] = links;
+                if (!string.IsNullOrEmpty(self) && relationship.LinkType.HasFlag(LinkType.Self))
```
joukevandermaas: I think for resources, we should both check their resources `LinkType` as well as the relationship's `LinkType`. It should probably resolve to the most restricting value.

Exampe:

```
PersonResource: WithLinks(All)
  - BelongsTo<CompanyResource> Job, WithLinks(None)
CompanyResource: WithLinks(Self)
```

Now when serializing the `Job` relationship, what links should show up? If I read this code correctly, currently it will serialize the `self` link even if the `BelongsTo` specified not to.

We probably also need a test case for this.

Saule/Serialization/ResourceSerializer.cs
```diff
@@ -235,24 +248,39 @@ private JObject SerializeRelationships(ResourceGraphNode node)
...
+                    && relationship.Kind == RelationshipKind.BelongsTo
                     && kv.Value.SourceObject != null)
                 {
                     relationshipId = (string)data["id"];
                 }
 
                 var links = new JObject();
-                AddUrl(links, "self", _urlBuilder.BuildRelationshipPath(node.Resource, node.Key.Id.ToString(), kv.Value.Relationship));
-                AddUrl(links, "related", _urlBuilder.BuildRelationshipPath(node.Resource, node.Key.Id.ToString(), kv.Value.Relationship, relationshipId));
+                var self = _urlBuilder.BuildRelationshipPath(node.Resource, node.Key.Id.ToString(), relationship);
```
joukevandermaas: We don't need to generate these if we might not use them, I think they should go inside the `if`s below.

bjornharrtell: Needed because I check for empty link generation. I.e unrelated to the LinkType configuration I want to avoid generating the links if the UrlBuilder returns null which I think is in line what you once wrote in https://github.com/joukevandermaas/saule/issues/121?

Saule/Serialization/ResourceSerializer.cs
bjornharrtell: Good question, I have a hard time getting a grip on what would be correct/desired behaviour here. Will have to think about it some more.

joukevandermaas: I wrote down some thoughts on this below. Please let me know what you think.

It's most likely that one of the two will just be `All`. I expect the most common scenario is to just override something in one place (i.e. either the relationship or on the model, depending which best maps to the task.

This implies to me that we should take the *most restrictive* of the options. Examples:

- Person -> Company (Job relationship). Person and company both have link type `All`, the relationship has link type `Related`. When serializing individual person or company records, we should serialize all links. When serializing the job relationship, we should serialize only `related` links.
- Person -> Company (Job relationship). Person has link type `All`, company has `Related`. The relationship has link type `All`. When serializing individual person records, we should serialize all links. For company records, we should not serialize anything (because there's no related links in that case). For the job relationship, we should serialize only `related` links.

One case where I'm not sure:
- Person -> Company (Job relationship). Person and company both have link type `None`. The relationship has link type `Related`. When serializing individual records, we should serialize no links. For the job relationship, what should we do? It sort of makes sense to serialize no links, but then again, the user explicitly set the link type to `Related` so they probably want those.

I guess the solution would be something like this:

1. When not serializing relationships, there is only one option so this should be simple.
2. Otherwise, if one of the options is `All`, take the other option.
3. In all other cases, use the logical 'or' of the options.

Examples of this:

- `All` and `None` -> serialize none
- `All` and `Related` -> serialize related
- `Related` and `None` -> serialize none
- `Related` and `Self` -> serialize related & self only (where they make sense)

Of course, we always have the option to just ignore one of the two:

1. When serializing the `related` hash, use the relationship link type
2. When serializing models (i.e. the top level `data` or `included` hash), use the resource link type

This is simple (I believe this is what the code in this PR does now) and seems the easiest to document.

bjornharrtell: I'm leaning on keeping it simple. The more complex path still hurts my brain. If one has other needs it could be implemented as a custom UrlBuilder. That said I might be shooting myself in the foot here. What is your opinion?

joukevandermaas: Yeah, I tend to agree. We can always expand on it later.

joukevandermaas:
I like the changes! I think this can be merged once the documentation has been updated (I think `docs/generating-links.md` is a good spot).

bjornharrtell:
Great, will add `docs/generating-links.md` asap. But what about `RelatedCanonical`? Shall I remove that one?

bjornharrtell:
Added documentation and tentatively removed `RelatedCanonical` enum type.

docs/Generating-links.md
```diff
@@ -7,8 +7,32 @@
...
+By using the enumeration `LinkType` you can opt out of link generation for a
+specific `ApiResource` or `Relationship` by setting it as in the following examples:
+
+In your `ApiResource` constructor:
+```csharp
+WithLinks(LinkType.None);
+```
+
+When defining a `Relationship`:
+```csharp
```
joukevandermaas: I think there needs to be a new line before a code block or the parser messes up the layout.

docs/Generating-links.md
```diff
@@ -7,8 +7,32 @@
...
+
+There are two complementary methods to control the link generation.
+
+### Resource and relationship configuration
+
+By using the enumeration `LinkType` you can opt out of link generation for a
+specific `ApiResource` or `Relationship` by setting it as in the following examples:
+
+In your `ApiResource` constructor:
+```csharp
```
joukevandermaas: I think there needs to be a new line before a code block or the parser messes up the layout.

joukevandermaas:
Thanks for seeing this through! It turned out really nice, I think.

PR #155, opened 2017-05-20
Fix regression missing 'data' key for relationships when using DisableDefaultIncluded

bjornharrtell:
Initially this PR added a test that fails on master but passes in stable to demonstrate the issue. The issue only reveals itself when DisableDefaultIncluded is used.

Now a fix is also included in this PR. It includes relationship data in the graph when using `DisableDefaultIncluded`, adds information of the parent property name for nodes which are children. It then uses this information when serializing the `includes` value.

Ping @laurence79.

### DIFF
--- a/Saule/Serialization/ResourceGraph.cs
+++ b/Saule/Serialization/ResourceGraph.cs
@@ -51,7 +51,8 @@ namespace Saule.Serialization
             object obj,
             ApiResource resource,
             ResourceGraphPathSet includePaths,
-            int depth)
+            int depth,
+            string propertyName = null)
         {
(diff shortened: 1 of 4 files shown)


bjornharrtell:
@laurence79 and @joukevandermaas the "Naive fix" I propose here is probably not the correct way to properly fix this. But it makes it pretty clear that the problem is that the graph is throwing away what is needed to produce the `data` elements.

bjornharrtell:
Worked on this some more and have added a fix that I think might be acceptable.

joukevandermaas:
👍 thanks for this!

Issue #159, opened 2017-06-12
Configurable inclusion for relationship object


bjornharrtell:
As described by http://jsonapi.org/format/#document-resource-object-relationships a "relationship object" MUST contain at least one of `links`, `data` or `meta`. When desired to lazy load related data it makes sense to only include the `links` key and omit `data`. This does not appear current possible with Saule so I propose to make that configurable.

Somewhat related to https://github.com/joukevandermaas/saule/issues/121.

joukevandermaas:
I think this is supported (although maybe not in the way you expect):

1. If there is a property for the relationship on the object, and the value is not `null`, include it
2. If there is a property for the relationship and the value is `null`, serialize `data: null`.
3. If the property for the relationship is not there, serialize only `links`.

I think there are tests that verify this behavior.

I think that when the user sets `DisableDefaultIncluded` and the user does not include the resource, the `data` hash should be omitted in favor of a `links` hash. I would prefer not to add any configuration or settings beyond that.

bjornharrtell:
I cannot verify that behaviour. At least for case 3 it seems I get `data: []` (regardless of `DisableDefaultIncluded`). Perhaps there is only a need for a fix for when using `DisableDefaultIncluded`.

joukevandermaas:
Sorry, the last paragraph refers to how I think it *should* work. I can see now that I was not entirely clear :smile:

bjornharrtell:
Ah :) Trying to fix that I find that it's a problem that a `Resource` class is not strictly related to the model it represents so I can't see if it has a property that represents the relationship. Have thought about that earlier but it had no real impact until now... to me it would perhaps make sense to point a Resource to a Model type, but that brings up the idea to bring Resource and Model closer in other ways (auto-defining attributes from properties etc) and would rather not go into that now.

bjornharrtell:
FYI will continue to try to find the easiest way to fix no 3.

bjornharrtell:
I think I've found a fix for no 3 in #161 but it's currently failing the test case "Handles recursive properties on resource objects" where I'm not sure if the old behaviour should/need to be reproduced as is by making the check recursive.

bjornharrtell:
Looked at the test case "Handles recursive properties on resource objects" again and can't make sense of it really, at least in the context of how it relates to resource inclusion. :( @joukevandermaas can you shed some light?

bjornharrtell:
[...]

PR #160, opened 2017-06-12
Fix test name typo

bjornharrtell:


joukevandermaas:
Ha, I guess double negatives can be confusing ;-)

I wonder if this test is correct given #159? In any case I will merge this (Appveyor seems to have choked on the build).

PR #161, opened 2017-06-12
Omit data for relationship objects not existing as property in the original model

bjornharrtell:
Related to https://github.com/joukevandermaas/saule/issues/159.

### DIFF
--- a/Saule/Serialization/ResourceSerializer.cs
+++ b/Saule/Serialization/ResourceSerializer.cs
@@ -293,7 +293,7 @@ namespace Saule.Serialization
 
                 var item = new JObject();
 
-                var data = SerializeRelationshipData(kv.Value);
+                var data = SerializeRelationshipData(node, kv.Value);
 
                 var relationshipId = default(string);
 
@@ -334,7 +334,7 @@ namespace Saule.Serialization
             return response;
         }
 
-        private JToken SerializeRelationshipData(ResourceGraphRelationship relationship)
+        private JToken SerializeRelationshipData(ResourceGraphNode node, ResourceGraphRelationship relationship)
         {
             // short circuit if not included in graph
             if (!relationship.Included)
@@ -344,9 +344,7 @@ namespace Saule.Serialization
 
             // check if the relationship property exists on the underlying model and if not bail with null
             // NOTE: this logic refers to https://github.com/joukevandermaas/saule/issues/159
-            // TODO: does not detect recursive properties on resource objects failing test "Handles recursive properties on resource objects"
-            var modelType = _value.GetType().GetGenericTypeParameterOfCollection() ?? _value.GetType();
-            if (modelType.GetProperty(relationship.Relationship.PropertyName) == null)
+            if (node.SourceObject.GetType().GetProperty(relationship.Relationship.PropertyName) == null)
             {
                 return null;
             }
--- a/Saule/Serialization/ResourceSerializer.cs
+++ b/Saule/Serialization/ResourceSerializer.cs
@@ -336,7 +336,17 @@ namespace Saule.Serialization
 
         private JToken SerializeRelationshipData(ResourceGraphRelationship relationship)
         {
+            // short circuit if not included in graph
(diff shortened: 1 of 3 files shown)


bjornharrtell:
Rebased and cleaned up. How to handle the remaining test case failure for "Handles recursive properties on resource objects" still eludes me...

bjornharrtell:
In my opinion the problem is that an `ApiResource` has no explicit connection to the underlying model. I can deduce it for the requested root `ApiResource` as the serializer knows about it (`_value`) but for recursive properties there is no such possibility.

To me it would make sense to have an explicit connection beteween `ApiResource` and underlying model but not sure if/how that could be added to Saule at this point.

I'll keep at it but would appreciate you opinion on this @joukevandermaas.

bjornharrtell:
Ah! I was mistaken, the information did exist in the ResourceGraphNode instance. Not entirely sure how it's wired up but it seems to do the trick so now this PR is ready for review. :)

joukevandermaas:
Glad you managed to get it working. I'll try to review asap.

FWIW I see it as a benefit of Saule that the resources (public API of your server) are decoupled from the models (private API of your server) a bit. It allows you to serialize different objects into the same resource. That might seem crazy, but it's actually really convenient if you use CQRS. For an example of this, see [this blog post](https://lostechies.com/jimmybogard/2015/05/05/cqrs-with-mediatr-and-automapper/). 

In any case if your app is different and only uses one model for any particular resource, you can already build a generic `ApiResource` using only public API:

```csharp
public class MyApiResource<T> : ApiResource
{
    public MyApiResource()
    {
        var props = typeof(T).GetProperties();

        var attrs = props.Where(p => p.GetFirstAttributeOrDefault<AttrAttribute>() != null);
        var belongsTo = props.Where(p => p.GetFirstAttributeOrDefault<BelongsToAttribute>() != null);
        var hasMany = props.Where(p => p.GetFirstAttributeOrDefault<HasManyAttribute>() != null);

        // these are generic, so we need to use reflection for this
        var belongsToMethod = GetType().GetMethods()
                  .Single(m => m.GetParameters().Length == 1 && m.Name == nameof(BelongsTo))
                  .GetGenericMethodDefinition();
        var hasManyMethod = GetType().GetMethods()
                  .Single(m => m.GetParameters().Length == 1 && m.Name == nameof(HasMany))
                  .GetGenericMethodDefinition();
        var genericThisType = GetType().GetGenericTypeDefinition();

        foreach (var prop in attrs)
        {
            Attribute(prop.Name);
        }

        foreach (var prop in belongsTo)
        {
            var genericType = genericThisType.MakeGenericType(prop.PropertyType);
            belongsToMethod.MakeGenericMethod(genericType).Invoke(this, new object [] { prop.Name });
        }

        foreach (var prop in hasMany)
        {
            var genericType = genericThisType.MakeGenericType(prop.PropertyType.GetGenericTypeParameterOfCollection());
            hasManyMethod.MakeGenericMethod(genericType).Invoke(this, new object[] { prop.Name });
        }
    }
}
```

You can use in controllers like this:

```csharp
[HttpGet]
[ReturnsResource(typeof(MyApiResource<MyModel>))]
public IEnumerable<MyModel> GetMyModel()
{
    return GetMyModels();
}
```

for models like this:

```csharp
public class MyModel
{
    [Attr]
    public string MyName { get; set; }

    [BelongsTo]
    public MyOtherModel OtherThing { get; set; }

    [HasMany]
    public IEnumerable<MyOtherModel> MoreOtherThings { get; set; }
}
```

I didn't test this code but it should work :smile:. I could see a potential future where this is the default behavior for Saule if you do not specify a `ReturnsResource` attribute, but I'm generally not a big fan of adding serialization-specific attributes to my domain models. As long as it is possible to do this, I'm not sure the code needs to be part of Saule itself.

bjornharrtell:
Thanks the detailed answer, very informative and after reading it I think I agree on your view of decoupling. :)

joukevandermaas:
Sorry for the late review. Thanks!

PR #163, opened 2017-06-12
Refactor collection type extractor into TypeExtensions

bjornharrtell:
Also avoids reflection on type name.

### DIFF
--- a/Saule/Serialization/ResourceSerializer.cs
+++ b/Saule/Serialization/ResourceSerializer.cs
@@ -90,10 +90,7 @@ namespace Saule.Serialization
             if (typeof(IEnumerable).IsAssignableFrom(valueType))
             {
                 isCollection = true;
-
-                // if our value is an enumerable, get the type of the items in the collection
-                var collectionType = valueType.GetInterfaces().FirstOrDefault(i => i.Name.Contains("IEnumerable") && i.IsGenericType);
-                valueType = collectionType?.GenericTypeArguments[0] ?? valueType;
+                valueType = valueType.TryGetCollectionType();
             }
 
             var metaObject = _resource.GetMetadata(_value, valueType, isCollection);
--- a/Saule/Serialization/ResourceSerializer.cs
+++ b/Saule/Serialization/ResourceSerializer.cs
@@ -90,7 +90,7 @@ namespace Saule.Serialization
             if (typeof(IEnumerable).IsAssignableFrom(valueType))
             {
                 isCollection = true;
-                valueType = valueType.TryGetCollectionType();
+                valueType = valueType.GetGenericTypeParameterOfCollection() ?? valueType;
             }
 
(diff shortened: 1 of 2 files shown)


Saule/TypeExtensions.cs
```diff
@@ -33,5 +33,21 @@ public static IEnumerable<Type> GetInheritanceChain(this Type type)
...
+        {
+            return type.IsGenericType && type.GetGenericTypeDefinition() == typeof(IEnumerable<>);
+        }
+
+        /// <summary>
+        /// Get the type of the items if the type is an <see cref="IEnumerable{T}"/> (if not it will return the input type).
+        /// </summary>
+        /// <param name="type">The type to apply this extension method to.</param>
+        /// <returns>The type of the items if the type is an <see cref="IEnumerable{T}"/> (if not it will return the input type).</returns>
+        public static Type TryGetCollectionType(this Type type)
```
joukevandermaas: This name is a bit confusing; maybe `GetGenericTypeParameterOfCollection` is more clear?

Saule/TypeExtensions.cs
```diff
@@ -33,5 +33,21 @@ public static IEnumerable<Type> GetInheritanceChain(this Type type)
...
+        {
+            var collectionType = type.GetInterfaces().FirstOrDefault(i => i.IsEnumerable());
+            return collectionType?.GenericTypeArguments[0] ?? type;
```
joukevandermaas: I'd prefer if this just returned `null` if the type could not be found. Now the behavior is really confusing from the name (`TryX` implies returning a `bool` and setting an `out` parameter to me).

Saule/Serialization/ResourceSerializer.cs
```diff
@@ -90,10 +90,7 @@ private JToken SerializeMetadata()
             if (typeof(IEnumerable).IsAssignableFrom(valueType))
             {
                 isCollection = true;
-
-                // if our value is an enumerable, get the type of the items in the collection
-                var collectionType = valueType.GetInterfaces().FirstOrDefault(i => i.Name.Contains("IEnumerable") && i.IsGenericType);
-                valueType = collectionType?.GenericTypeArguments[0] ?? valueType;
+                valueType = valueType.TryGetCollectionType();
```
joukevandermaas: If `GetGenericTypeParameterOfCollection` would return `null` if `valueType` was not a generic collection, this line could be

```csharp
valueType = valueType.GetGenericTypeParameterOfCollection() ?? valueType;
```

which would be more clear if you didn't know the context.

joukevandermaas:
Why is this needed? This code is only used in one spot as far as I can tell?

bjornharrtell:
Agreed on all review points, should be fixed now and rebased against master.

joukevandermaas:
Awesome, thanks!

Issue #128, opened 2016-08-10
Accept page size from pagination query string


bxh:
It's common for a RESTful client to suggest a page such as `?page[number]=0&page[size]=20`. Saule presently can only specify page size on the server side. Can we add support to `page[size]`?

erikhejl:
I'm taking on this work in a fork of ours, but I did have some questions.  I'm puzzled as to why QueryContext properties are assigned both by the Filter/Sorting/Pagination attributes and also by the JsonApiSerializerOfT.   It seems like the latter would be problematic?

joukevandermaas:
@erikhejl `JsonApiSerializer<T>` is can be used in setup. This is relevant for filtering, because you can provide custom filtering logic. I doubt this is relevant to pagination, I think you will only need to care about the non-generic `JsonApiSerializer`.

PR #165, opened 2017-07-11
Added pagination page size

erikhejl:
This pull request is for issue #128 (https://github.com/joukevandermaas/saule/issues/128#issuecomment-312198306).  Default values for page size and page size limits can be configured imperatively at app start with the JsonApiConfiguration.PaginationConfig property, or with web.config appSetting values.  See the PaginationConfig PageSizeAppSettingKey and PageSizeLimitKey constants.  The Pagination attribute has had a PageSizeLimit property added for overriding the application default values.

I hold the opinion that with page[size] from the query implemented, that the default of 10 as it existed with the Pagination attribute should be done away with, and that the library should require both page[number] and page[size] if either is supplied, but this implementation has the page size default set to 10 for compatibility with legacy versions of Saule.

This is my first time contributing to any open source projects, so my apologies in advance if there are issues with this pull request.

### DIFF
--- a/Saule/Http/PaginatedAttribute.cs
+++ b/Saule/Http/PaginatedAttribute.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Linq;
+using System.Net;
 using System.Net.Http;
 using System.Web.Http.Controllers;
 using System.Web.Http.Filters;
@@ -15,7 +16,8 @@ namespace Saule.Http
     [AttributeUsage(AttributeTargets.Method)]
     public sealed class PaginatedAttribute : ActionFilterAttribute
     {
-        private int _perPage = 10;
+        private int? _perPage;
+        private int? _queryPageSizeLimit;
 
         /// <summary>
         /// Gets or sets the number of items to return per response.
@@ -24,7 +26,7 @@ namespace Saule.Http
         {
             get
             {
-                return _perPage;
+                return _perPage ?? Constants.QueryValues.ValueNotSpecified;
             }
 
             set
@@ -34,7 +36,38 @@ namespace Saule.Http
                     throw new ArgumentOutOfRangeException(nameof(PerPage), value, "Must have at least one item per page.");
                 }
 
+                if (_queryPageSizeLimit.HasValue && value > _queryPageSizeLimit)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(PageSizeLimit), value, "Items per page cannot be larger than the page size limit.");
+                }
+
                 _perPage = value;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the maximum page size to accept from a URL query string.
+        /// </summary>
+        public int PageSizeLimit
+        {
+            get
+            {
+                return _queryPageSizeLimit ?? Constants.QueryValues.ValueNotSpecified;
+            }
+
+            set
+            {
+                if (value < 1)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(PageSizeLimit), value, "Must have at least one item per page.");
+                }
+
+                if (_perPage.HasValue && value < _perPage)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(PageSizeLimit), value, "PageSizeLimit cannot be smaller than page size.");
+                }
+
+                _queryPageSizeLimit = value;
             }
         }
 
@@ -44,13 +77,14 @@ namespace Saule.Http
         /// <param name="actionContext">The action context.</param>
         public override void OnActionExecuting(HttpActionContext actionContext)
         {
-            var context = new PaginationContext(
+            var paginationContext = new PaginationContext(
                 actionContext.Request.GetQueryNameValuePairs(),
-                PerPage);
+                _perPage,
+                _queryPageSizeLimit);
 
             var query = GetQueryContext(actionContext);
 
-            query.Pagination = context;
+            query.Pagination = paginationContext;
             base.OnActionExecuting(actionContext);
         }
 
--- a/Saule/Http/PaginationConfig.cs
+++ b/Saule/Http/PaginationConfig.cs
@@ -8,19 +8,6 @@ namespace Saule.Http
     public class PaginationConfig
     {
         /// <summary>
-        /// Web.Config appSettings key for the default page size.
-        /// </summary>
-        public const string PageSizeAppSettingKey = "JsonApi.PageSizeDefault";
-
-        /// <summary>
-        /// Web.Config appSettings key for the default limit for page[size] query parameter values.
-        /// </summary>
-        /// <remarks>
-        /// Omitting this value from configuration will result in the page[size] query parameter being ignored.
-        /// </remarks>
-        public const string PageSizeLimitKey = "JsonApi.PageSizeLimit";
-
-        /// <summary>
         /// Initializes a new instance of the <see cref="PaginationConfig"/> class.
         /// </summary>
         public PaginationConfig()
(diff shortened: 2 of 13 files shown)


Tests/Tests.csproj
```diff
@@ -28,6 +28,7 @@
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <NoWarn>CS0618</NoWarn>
```
joukevandermaas: What is this for?

Saule/StringDictionaryExtensions.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Saule
+{
+    /// <summary>
+    /// Extension methods for operating on IDictionary[string, string]
+    /// </summary>
+    public static class StringDictionaryExtensions
```
joukevandermaas: This should be internal?

Saule/Http/PaginatedAttribute.cs
```diff
@@ -34,23 +36,55 @@ public int PerPage
...
+                }
+
                 _perPage = value;
             }
         }
 
         /// <summary>
+        /// Gets or sets the maximum page size to accept from a URL query string.
+        /// </summary>
+        public int PageSizeLimit
```
joukevandermaas: Perhaps this should be `int?` instead of using `-1` to represent unlimited size? I'm curious about the reason you went for `int` here, but `int?` in the config.

Saule/Http/PaginationConfig.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            // TODO: for compatibility with legacy code only.  Recommend doing away with this in favor of app level configuration
+            DefaultPageSize = 10;
+        }
+
+        /// <summary>
+        /// Gets or sets default page size to use for query actions decorated
+        /// with a [Paginated] attribute that does not specify a PerPage value.
+        /// </summary>
+        public int DefaultPageSize { get; set; }
```
joukevandermaas: Let's just set this to 10 and not make it configurable; if people want a different page size they can change it using the `PaginatedAttribute`.

Saule/Http/PaginationConfig.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        /// <summary>
+        /// Gets or sets the default value to use for the maximum allowable page
+        /// size to accept from the client with a [Paginated] attribute that does
+        /// not specify a MaxClientPageSize value.
+        /// </summary>
+        public int? DefaultPageSizeLimit { get; set; }
```
joukevandermaas: Let's just set this to 100 and not make it configurable; if people want a different page size they can change it using the `PaginatedAttribute`.

Saule/Http/PaginationConfig.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        /// <summary>
+        /// Gets a value indicating whether or not page[size] query string parameters will be honored.
+        /// </summary>
+        public bool AllowQueryPageSize
+        {
+            get { return DefaultPageSizeLimit.HasValue; }
```
joukevandermaas: This should just always be `true`, I think. I know it's technically a breaking change but I doubt anyone will mind (in any case they can set the limit).

Saule/Http/PaginationConfig.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        /// <summary>
+        /// Create pagination configuration from web.config app settings.
+        /// </summary>
+        /// <returns>PaginationConfig instance with values from web.config.</returns>
+        public static PaginationConfig FromWebConfig()
```
joukevandermaas: I don't understand the need for the web.config configuration. I get that it may be convenient sometimes, but it feels like a lot of extra stuff to think about within Saule. For example, why put pagination configuration in there, but not other things?

I would prefer we allow this configuration through code (at runtime) only.

Saule/Http/PaginationConfig.cs
```diff
@@ -0,0 +1,77 @@
+﻿using System.Configuration;
+
+namespace Saule.Http
+{
+    /// <summary>
+    /// Pagination configuration.  Property values can be set programmatically or be read from web.config app settings.
+    /// </summary>
+    public class PaginationConfig
```
joukevandermaas: I'm somewhat questioning the need for this configuration. Perhaps it's good enough to pick some sensible defaults instead? We can always add it again later if there's a compelling reason.

Tests/Tests.csproj
erikhejl: The Obsolete takes you placed on the JsonApiMediaTypeFormatter generated compiler warnings in the unit tests you made for them.  This suppresses those warnings in the unit test project and leaves a cleaner build.  It's possible this could be done at the file level, this was just readily available in the VS project settings.

Saule/Http/PaginatedAttribute.cs
erikhejl: I tried that initially.  Int isn't a primitive and thus cannot be used as an attribute property if you want to initialize it within the actual decorator.

Saule/StringDictionaryExtensions.cs
erikhejl: Sure, I'll change it.

Saule/Http/PaginationConfig.cs
erikhejl: I disagree with the notion of the sensible defaults.  There's no reason to limit options for consumers of Saule, and case in point, I added this because a default of 10 wasn't appropriate for my use case.

Saule/Http/PaginationConfig.cs
erikhejl: I put this in as I found it to be a bit reckless to allow consumers of our APIs to request page sizes as large as they want.  The likely outcome is we'll end up with clients who make huge requests because it simplifies things on their end, meanwhile our servers are being beaten to a pulp.  Thus I don't think client/query page sizes should be accepted unless the application has a defined limit.

Saule/Http/PaginationConfig.cs
erikhejl: My feelings on this are the same as [above](https://github.com/joukevandermaas/saule/pull/165#discussion_r127592303).

Saule/Http/PaginationConfig.cs
erikhejl: My feelings on this are the same as [above](https://github.com/joukevandermaas/saule/pull/165#discussion_r127592303).

Saule/Http/PaginationConfig.cs
erikhejl: This is fine.  It was a minor convenience, but it only saves a couple of lines of app start code.  I'll remove it.

erikhejl:
Regarding the general remarks:
"I think there should be some more edge-case tests for this. You can add them in the same file you added the current ones."
I'm happy to do this but I'm not sure what sort of situations you have in mind.

"I've tried to keep Saule pretty low on configuration wherever possible. Any time we can pick sensible defaults instead, I prefer that option."
We've got differing views on this when it comes to the pagination, but I did make it so that the library works configuration-free with the same behavior as it had before.

Saule/Http/PaginationConfig.cs
erikhejl: Actually, this is something I put in place.  The concern was that allowing client to specify the page size with no limit defined is puts you at risk for clients being greedy to simplify things on their end and putting an excessive load on the servers as a result.  While I still think it's a terrible idea to stand up an app without a limit configured, I decided to remove this and not limit choice in applications that are using Saule.

joukevandermaas:
Some test scenarios I could think of:

- set maximum page size to 1, assert that:
  - client-specified page size of 1 is allowed and works
  - client-specified page size of > 1 is rejected
  - no client-specified page size works (and returns one result)
- set maximum page size to 'infinity' (I believe this is `-1` in the code?), assert that:
  - client-specified page size of 1 is allowed and works
  - client-specified page size of 1000 is allowed and works
  - no client-specified page size works and returns 10 results (or whatever the default is)
- set maximum page size to 50, assert that:
  - client-specified page size of 1 is allowed and works
  - client-specified page size of 100 is rejected
  - no client-specified page size works and returns 10 results (or whatever the default is)

If you can think of some more edge cases based on the code, feel free to add tests there too.

---

As for the configuration discussion, I don't really like the complexity 'default' configuration adds. It just adds one other thing to keep in mind with everything we implement, which is not a good trade-off at the moment, since I'm the only maintainer of this project. The option to have many people of different experience levels contribute to Saule has been great, and I would like to keep it that way. 

In general I believe most people have roughly the same needs, and default values can capture that. This is a more-and-more widely accepted view in the web ecosystem. Look for example at Ruby on Rails, the new ASP Core, Ember, create-react-app, etc. 

Configuration is sometimes necessary, but I often prefer to capture those needs in an abstraction instead. For example, for pagination we could make the pagination logic public and extensible (like we did with url generation or filtering). Then consumers of Saule could implement their own pagination schemes and change any defaults they want that way. That tends to be a a lot harder to do if we have many configurable points already, because custom extensions would also need access to those, complicating their API.

Convenience is a compelling argument in many cases, but in this case I don't believe so. In my app, I would much rather find the allowed page sizes near the endpoint's action method than in some global configuration file. That way, if I'm not sure, it's much easier to find.

In any case, thanks for your responses and continued effort on this PR. I hope you can see my point of view :smile:

sergey-litvinov-work:
Hey guys, just want to check this pull request. We also need similar functionality, but as i see there is no activity here for one month. Maybe help is needed with changes to this pull request?

joukevandermaas:
@sergey-litvinov-work I'm pretty happy with the PR as-is, except for two small things:

- remove the `PaginationConfig` class
- add the tests as described in the above scenario

edward-rosado:
@joukevandermaas we actually have a use case where we want to have a default configuration to be different for specific applications. In 1 case we want the default to be 10, but in another case we want the default to be 50. We dont want to force the caller to pass in a pagesize of 50 when literally every call will need to return 50. I see value in leaving this particular configuration, however i do agree with the general assesment on having as small configurations as possible.

Would love it if we could leave it in, but also get better unit tests for the cases you specified. I can work with my team to get that done for you.

Issue #167, opened 2017-07-25
I can't call method for upload Files


tomasjurasek:
Hi,
I have classic method for upload files.
` public async Task<IHttpActionResult> UploadFile(){......}`
[UploadFile](https://docs.microsoft.com/en-us/aspnet/web-api/overview/advanced/sending-html-form-data-part-2)

But when I use [ReturnsResource(typeof(GlobalFileModelResource))] doesnt works. 
![image](https://user-images.githubusercontent.com/1623933/28591686-dca7de98-7186-11e7-84c4-efb82f14f44d.png)

It is some specific settings for IsMimeMultipartContent?

Thanks
Tomas

joukevandermaas:
I think if you use the following overload for `ConfigureJsonApi` it might work:

```csharp
ConfigureJsonApi(
    jsonApiConfiguration: new JsonApiConfiguration(),
    overwriteOtherFormatters: false)
```

If not, that's a bug (in my opinion). Note that to my knowledge, the JSON API spec itself does not mention uploading files and/or multipart content.

tomasjurasek:
Does not works with these options . The same errors.

tomasjurasek:
Im playing with this issue. I think we need overload this method

```
       public static void ConfigureJsonApi(
            this HttpConfiguration config,
            JsonApiConfiguration jsonApiConfiguration,
            bool overwriteOtherFormatters)
        {
            config.MessageHandlers.Add(new PreprocessingDelegatingHandler(jsonApiConfiguration));
            var formatter = new JsonApiMediaTypeFormatter(jsonApiConfiguration);

            if (overwriteOtherFormatters)
            {
                config.Formatters.Clear();
                config.Formatters.Add(formatter);
            }
            else
            {
                config.Formatters.Insert(0, formatter);
            }
        }
```

Because we dont want clear all formatters and insert only one. We need add formatters to end of collection when we want use WEB API together with JSON API. When we rework this method it will be breaking change, overload is better solution.

@joukevandermaas

PR #168, opened 2017-08-02
configureJsonApi overload (formatter can be only add at end of the collection, without delete other formatters

tomasjurasek:
I added more options for configure formatters in ConfigureJsonApi. For example if want combinate JSON API with WebAPI together.
#167

### DIFF
--- a/Saule/ConfigureFormattersEnum.cs
+++ b/Saule/ConfigureFormattersEnum.cs
@@ -1,0 +1,29 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Saule
+{
+    /// <summary>
+    /// Configuration enum for formatters
+    /// </summary>
+    public enum ConfigureFormattersEnum
+    {
+        /// <summary>
+        /// Formatter will be inserted at the start of the collection
+        /// </summary>
+        AddFormatterToStart = 0,
+
+        /// <summary>
+        /// Formatter will be inserted at the end of the collection
+        /// </summary>
+        AddFormatterToEnd = 1,
+
+        /// <summary>
+        /// Other formatters will be cleared
+        /// </summary>
+        OverwriteOtherFormatters = 2,
+    }
+}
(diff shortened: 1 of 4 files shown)


Saule/ConfigureFormattersEnum.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Saule
+{
+    /// <summary>
+    /// Configuration enum for formatters
+    /// </summary>
+    public enum ConfigureFormattersEnum
```
joukevandermaas: Maybe `FormatterPriority`?

Saule/ConfigureFormattersEnum.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Saule
+{
+    /// <summary>
+    /// Configuration enum for formatters
```
joukevandermaas: This description isn't very clear. Maybe describe the purpose rather than what it does?

Saule/Http/HttpConfigExtensions.cs
```diff
@@ -48,15 +48,44 @@ public static void ConfigureJsonApi(this HttpConfiguration config, JsonApiConfig
...
+            }
+        }
+
+        /// <summary>
+        ///  Sets up serialization and deserialization of Json Api resources.
+        /// </summary>
+        /// <param name="config">The <see cref="HttpConfiguration"/> that is used in the setup of the application.</param>
+        /// <param name="jsonApiConfiguration">Configuration parameters for Json Api serialization.</param>
+        /// <param name="configureFormatters">
+        /// AddFormatterToStart - Formatter will be inserted at the start of the collection.
```
joukevandermaas: This should be pretty clear from the enum values; maybe something like `Determines the relative position of the JSON API formatter`?

joukevandermaas:
Thanks for this! I will publish a pre-release shortly with these changes in it.

PR #172, opened 2017-09-04
Added functionality to return list of HttpErrors instead of just one error

sergey-litvinov-work:
Hi,

We have a case when we need to return list of errors. Like if we do some internal validation and there are more than one error, then we need to return all of them. But right now Saule supports just one HttpError to be returned.

As i see in the JsonApi spec - http://jsonapi.org/format/#errors - it supports multiply errors in the response.
So i did this change and pull request.

I updated `PreprocessResult` to have `List<ApiError>` instead of just one `ApiError` and updated code to work with list of errors. Also added integration test and unit test for `ErrorSerializer`.

Could you please review when you have time?
Thanks


sergey-litvinov-work:
@joukevandermaas added two more integration tests. One of them is for case when we return just one http error instead of list of errors, and another one when invalid operation exception happens.

joukevandermaas:
Perfect! Thanks!

I will publish a new pre-release with these changes.

sergey-litvinov-work:
Thanks!

Issue #169, opened 2017-08-16
Deserialize kebab-cased nested attributes


atkretsch:
Similar to #131, but for deserialization.

Given the following class structure:
```cs
public class Person : ApiResource
    {
        public Person()
        {
            Attribute("Name");
        }

        public Name Name { get; set; }
    }

    public class Name
    {
        public string FirstName { get; set; }
        public string LastName { get; set; }
    }
```

...the following JSON (e.g. in a `POST` request):
```json
{
    "data": {
        "attributes": {
            "name": {
                "first-name": "John",
                "last-name": "Doe"
            }
        }
    }
}
```

...will result in a `Person` class such that `person.Name.FirstName == null` and `person.Name.LastName == null`.

However, if the nested attributes are changed to UpperCamelCase (e.g. `FirstName` and `LastName`), then the result is correct (e.g. `person.Name.FirstName == "John"`, `person.Name.LastName == "Doe"`).  And when serializing (e.g. in a response), the result has the expected kebab-case (because of #131).

I know many examples of nested objects like this can probably be refactored as relationships, etc., but as discussed in #131 it's sometimes useful to use "value objects" for grouping of closely-related attributes. 
 Additionally, if the default behavior for serialization is to handle kebab-casing all the way down, then it seems reasonable that the same should be true for deserialization for consistency and request/response symmetry.

From poking around the code a bit, it seems like the `ResourceDeserializer` would have to be changed to recursively call `ToFlatStructure()` for non-primitive attribute values (similar to what it does for relationships).

joukevandermaas:
The solution in the serializer was to use this 'contract resolver': https://github.com/joukevandermaas/saule/blob/master/Saule/Serialization/JsonApiContractResolver.cs (see #132).

I'm guessing we could do the same for the deserializer.

atkretsch:
@joukevandermaas Yes, that makes more sense than my initial idea.  If you don't mind I'll take a crack at it and put up a PR.

joukevandermaas:
That would be great!

esotery:
Hello gentlemen, any update regarding this issue?:)

joukevandermaas:
@esotery I'm still very open to a PR!

atkretsch:
Unfortunately I haven't had time to take another look at this yet.  Hoping I will in the near future, but in the meantime, @esotery feel free to do so yourself if you're so inclined!

esotery:
[...]

PR #176, opened 2017-09-26
#169 Deserialize kebab-case

esotery:
Hi there,

fixed the issue. All unit tests are passing. I have added mine and verified the functionality but haven't actually pushed it in because the data and structure is kinda proprietary.

### DIFF
--- a/Saule/Serialization/JsonApiContractResolver.cs
+++ b/Saule/Serialization/JsonApiContractResolver.cs
@@ -1,22 +1,12 @@
-﻿using System;
-using System.Collections.Generic;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Serialization;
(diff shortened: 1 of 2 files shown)


joukevandermaas:

I think it would be nice to have one small unit test that verifies the new behavior, so we prevent regressions in the future. There should already be some models you can use in the Tests project.

joukevandermaas:
@PhyberApex yes, i seem to recall fixing this when I was writing some extra tests. I had forgotten about this PR so I didn't post a comment. Are you able to see if #169 is fixed for you in the latest pre-release?

PhyberApex:
Right now I am not able to check this. But if you want me to I can give it a go on monday and report back!

~Cheers

PR #179, opened 2017-11-17
Made ResourceDeserializer public

madsphi:
When testing APIs, it's super useful to be able to use `ResourceDeserializer` to deserialize json into specific objects.

We might for example receive a json string containing user data and nested data about the organization the user belongs to.

The json string can then be converted to a user object containing an organization object like this:

```
public T ConvertJson(string json) where T : class
{
	var jObject = JObject.Parse(json);
	var target = new ResourceDeserializer(jObject, typeof(T));
	return target.Deserialize() as T;
}
```

### DIFF
--- a/Saule/JsonApiSerializerOfT.cs
+++ b/Saule/JsonApiSerializerOfT.cs
@@ -97,7 +97,7 @@ namespace Saule
         /// <param name="object">Json to convert</param>
         /// <param name="type">Type to convert to</param>
         /// <returns>Json converted into the specified type of object</returns>
-        public dynamic Deserialize(JToken @object, Type type)
+        public object Deserialize(JToken @object, Type type)
         {
             var target = new ResourceDeserializer(@object, type);
             return target.Deserialize();
--- a/Saule/JsonApiSerializerOfT.cs
+++ b/Saule/JsonApiSerializerOfT.cs
@@ -91,6 +91,18 @@ namespace Saule
             return JsonApiSerializer.Serialize(preprocessResult);
         }
 
+        /// <summary>
+        /// Converts json into an object of a specified type
+        /// </summary>
+        /// <param name="object">Json to convert</param>
+        /// <param name="type">Type to convert to</param>
+        /// <returns>Json converted into the specified type of object</returns>
+        public dynamic Deserialize(JToken @object, Type type)
+        {
+            var target = new ResourceDeserializer(@object, type);
+            return target.Deserialize();
+        }
+
         private QueryContext GetQueryContext(IEnumerable<KeyValuePair<string, string>> filters)
         {
             var context = new QueryContext();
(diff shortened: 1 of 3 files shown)


joukevandermaas:

I understand that this is useful, but I don't think that this is the right solution. I'm more open to adding a `Deserialize` method to the `JsonSerializer<T>` class (which creates and calls the deserializer internally). The reason is that I want to be able to refactor deserialization (for example, it may need to have access to `ApiResource`s at some point) without having to worry about breaking public API.

If you are willing to make that change, please also add a test to make sure we don't break it later.

madsphi:
Ok. That makes sense.

`JsonSerializer<T>`, is that the `JsonApiSerializer<T>`? (I can't find any `JsonSerializer<T>` in the Saule solution)

joukevandermaas:
@madsphi Yes, sorry. I updated my previous comment to fix the typo.

Saule/JsonApiSerializerOfT.cs
```diff
@@ -91,6 +91,18 @@ public JToken Serialize(object @object, Uri requestUri)
             return JsonApiSerializer.Serialize(preprocessResult);
         }
 
+        /// <summary>
+        /// Converts json into an object of a specified type
+        /// </summary>
+        /// <param name="object">Json to convert</param>
+        /// <param name="type">Type to convert to</param>
+        /// <returns>Json converted into the specified type of object</returns>
+        public dynamic Deserialize(JToken @object, Type type)
```
joukevandermaas: Not a huge fan of `dynamic` here, but I'm assuming you have a good reason. I propose one of the following, but if that wouldn't work we can re-evaluate:

1. Change the signature to `public TResult Deserialize<T>(JToken)`
2. Change the signature to `public object Deserialize(JToken, Type)`

Alternatively we can have an overload for both.

madsphi: Thought about using option 2, but then I had to make a cast in the unit test/where I use the method.

i.e.
`var dsPerson = (Person)target.Deserialize(personJson, typeof(Person));
...dsPerson.FirstName;`

instead of just
`var dsPerson = target.Deserialize(personJson, typeof(Person));
...dsPerson.FirstName;`

That's why I preferred using dynamic.

joukevandermaas: It's fine by me to do a cast in the test, as long as it's likely that the consumers of the API will also do a cast. `dynamic` adds some overhead and feels a bit 'unconventional', so if that was the only reason I prefer not to do it.

madsphi: Ok. I committed a return `object` version.

joukevandermaas:
@madsphi Thanks a lot for seeing this through! I'll release a new pre-release with these changes in it.

madsphi:
Excellent - and thanks a lot for your great comments/suggestions.

PR #184, opened 2018-03-16
Fix regression with omitted data

bjornharrtell:
Regression was caused by https://github.com/joukevandermaas/saule/pull/180. When a one-to-one relationship is empty data would be omitted and that is invalid for a relationship object without links - and it would be unfortunate to be forced to include links for nullable one-to-one relationships.

### DIFF
--- a/Tests/Serialization/ResourceSerializerTests.cs
+++ b/Tests/Serialization/ResourceSerializerTests.cs
@@ -411,8 +411,8 @@ namespace Tests.Serialization
             Assert.NotNull(attributes["last-name"]);
             Assert.NotNull(attributes["age"]);
 
-            Assert.Equal(0, relationships["job"]["data"].Count());
-            Assert.Equal(0, relationships["friends"]["data"].Count());
+            Assert.Equal(JTokenType.Null, relationships["job"]["data"].Type);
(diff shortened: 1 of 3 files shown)


Tests/Serialization/ResourceSerializerTests.cs
```diff
@@ -392,8 +411,8 @@ public void HandlesNullValues()
             Assert.NotNull(attributes["last-name"]);
             Assert.NotNull(attributes["age"]);
 
-            Assert.Null(relationships["job"]["data"]);
-            Assert.Null(relationships["friends"]["data"]);
+            Assert.Equal(0, relationships["job"]["data"].Count());
```
joukevandermaas: `Assert.Null` in this context checks that the property isn't there at all? I think checking for count on a one-to-one relationship is confusing. In this case, if the relationship is `null`, I would expect the value to be `null`, so `relationships["job"]["data"].Type == JTokenType.Null`.

bjornharrtell: Agreed, reworked to be more intuitive.

joukevandermaas:
Awesome, thanks!

PR #185, opened 2018-03-17
Upgrade to stable deps

bjornharrtell:


joukevandermaas:
Generally love this but I'm worried about `JSON.Net` and `System.Http`. Are there any consequences for consuming projects?

bjornharrtell:
I honestly have no idea. :)

PR #187, opened 2018-04-04
opt into JsonApi per webapi endpoint

barsh:
For existing WebApi projects, that don't use JsonApi/Saule we need a way to preserve backwards compatibility and opt into JsonApi on a case by case basis.

For example: `api/v2/companies` uses the attribute filter `JsonApi` which configures it to return a JsonAPI response instead of the default response.

<kbd>image</kbd>

With this change, it's not necessary to call `ConfigureJsonApi` during application startup:

<kbd>image</kbd>

### DIFF
--- a/Saule/Http/JsonApiAttribute.cs
+++ b/Saule/Http/JsonApiAttribute.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Net.Http;
 using System.Web.Http.Filters;
 
 namespace Saule.Http
@@ -7,7 +8,7 @@ namespace Saule.Http
     /// An optional attribute that can be used to opt an api into returning a JsonApi response.
     /// </summary>
     [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class)]
-    public class JsonApiAttribute : ActionFilterAttribute
+    public sealed class JsonApiAttribute : ActionFilterAttribute
     {
         /// <summary>
         /// See base class documentation.
@@ -15,7 +16,23 @@ namespace Saule.Http
         /// <param name="context">The action context.</param>
         public override void OnActionExecuted(HttpActionExecutedContext context)
         {
-            JsonApiProcessor.ProcessRequest(context);
+            var config = new JsonApiConfiguration();
+            JsonApiProcessor.ProcessRequest(context.Request, context.Response, config, requiresMediaType: false);
+
+            if (context.Exception != null)
+            {
+                return;
+            }
+
+            var responseContent = context.Response.Content as ObjectContent;
+            if (responseContent == null)
+            {
+                return;
+            }
+
+            var formatter = new JsonApiMediaTypeFormatter(context.Request, config);
+
+            context.Response.Content = new ObjectContent(responseContent.ObjectType, responseContent.Value, formatter);
         }
     }
 }
--- a/Saule/Http/JsonApiAttribute.cs
+++ b/Saule/Http/JsonApiAttribute.cs
@@ -31,7 +31,7 @@ namespace Saule.Http
             var formatter = new JsonApiMediaTypeFormatter(new JsonApiConfiguration()).GetPerRequestFormatterInstance(
                 typeof(string),
                 actionExecutedContext.Request,
-                new MediaTypeHeaderValue("application/vnd.api+json"));
+                new MediaTypeHeaderValue(Constants.MediaType));
 
             actionExecutedContext.Response.Content = new ObjectContent(content.ObjectType, content.Value, formatter);
         }
--- a/Saule/Http/JsonApiAttribute.cs
+++ b/Saule/Http/JsonApiAttribute.cs
@@ -1,0 +1,39 @@
+﻿using System;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Web.Http.Filters;
+
+namespace Saule.Http
+{
+    /// <summary>
+    /// An optional attribute that can be used to opt an api into returning a JsonApi response.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class)]
+    public sealed class JsonApiAttribute : ActionFilterAttribute
+    {
+        /// <summary>
+        /// See base class documentation.
+        /// </summary>
+        /// <param name="actionExecutedContext">The action context.</param>
+        public override void OnActionExecuted(HttpActionExecutedContext actionExecutedContext)
+        {
+            if (actionExecutedContext.Exception != null)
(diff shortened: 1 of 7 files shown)


joukevandermaas:
@barsh Thanks so much for contributing!

A few notes:

1. The way this is implemented now, it will not support all of Saule's features. A lot of that depends on the delegating handler that is also registered in the configure methods.
2. Saule *does* support a way of optionally processing only certain requests (at least in the current pre-release). You can use this overload of the configure method:

```cs
config.ConfigureJsonApi(new JsonApiConfiguration(), FormatterPriority.AddFormatterToEnd);
```

Now *only* requests with the accept request header value of `application/vnd.api+json` will be processed:

```
Accept: application/vnd.api+json
```

That effectively enables *clients* to opt in to JSON API, which may not be what you need. To enable the server to opt in for specific endpoints, I think we need a solution that at least does the same amount of processing that the regular setup does (i.e. supports all features). Ideally we would find a way to make it possible to implement this `JsonApiAttribute` as a user of Saule, without needing to add it to the library itself.

Perhaps a class `JsonApiProcessor` and a method with a signature like this:

```cs
void ProcessRequest(HttpRequestMessage request, HttpResponseMessage response)
```

which does the same thing as [the delegating handler](https://github.com/joukevandermaas/saule/blob/master/Saule/Http/PreprocessingDelegatingHandler.cs#L72-L92). Then when a user of Saule implements the `JsonApiAttribute` action filter, they can call this `ProcessRequest` method before using the media-type formatter.

What do you think?

barsh:
@joukevandermaas I'll have a go at it over the weekend and update this PR

Saule/Http/JsonApiProcessor.cs
```diff
@@ -0,0 +1,61 @@
...
+            var statusCode = (int)context.Response.StatusCode;
+            if (statusCode >= 400 && statusCode < 500)
+            {
+                // probably malformed request or not found
+                return;
+            }
+
+            var value = context.Response.Content as ObjectContent;
+
+            var config = new JsonApiConfiguration();
```
barsh: @joukevandermaas, I couldn't figure out how to get the configuration, so I created a new one.  Is that OK?  Is there a way to get the configuration?

Saule/Http/JsonApiProcessor.cs
```diff
@@ -0,0 +1,61 @@
...
+            var responseContent = context.Response.Content as ObjectContent;
+            if (responseContent == null)
+            {
+                return;
+            }
+
+            var formatter = new JsonApiMediaTypeFormatter(config).GetPerRequestFormatterInstance(
+                typeof(string),
+                context.Request,
+                new MediaTypeHeaderValue(Constants.MediaType));
```
barsh: @joukevandermaas this works but it feels hacky.  Is there a better way to get the formatter?

Saule/Http/JsonApiProcessor.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// <summary>
+    /// Processes JSON API responses
+    /// </summary>
+    public class JsonApiProcessor
+    {
+        /// <summary>
+        /// process request as JSON API response
+        /// </summary>
+        /// <param name="context">context</param>
+        public static void ProcessRequest(HttpActionExecutedContext context)
```
joukevandermaas: Could this take the `HttpRequestMessage` and `HttpResponseMessage` (or similar) as parameters? The `ActionExecutedContext` is very closely tied to action filters.

Saule/Http/JsonApiProcessor.cs
joukevandermaas: That's fine, perhaps we can also create an overload that takes it as a parameter.

Saule/Http/JsonApiProcessor.cs
joukevandermaas: We could use the constructor overload:

```csharp
internal JsonApiMediaTypeFormatter(HttpRequestMessage request, JsonApiConfiguration config)
```

That one is called internally by `GetPerRequestFormatterInstance`.

Saule/Http/JsonApiProcessor.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Net.Http.Headers;
+using System.Web.Http.Filters;
+using Saule.Serialization;
+
+namespace Saule.Http
+{
+    /// <summary>
+    /// Processes JSON API responses
+    /// </summary>
+    public class JsonApiProcessor
```
joukevandermaas: Perhaps this class should be static or sealed.

Saule/Http/JsonApiAttribute.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Web.Http.Filters;
+
+namespace Saule.Http
+{
+    /// <summary>
+    /// An optional attribute that can be used to opt an api into returning a JsonApi response.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class)]
+    public class JsonApiAttribute : ActionFilterAttribute
```
joukevandermaas: Should this still be sealed?

Saule/Http/JsonApiProcessor.cs
joukevandermaas: Do you think we can call this from the delegating handler? Now the code is duplicated. I think since this is public, it would be awesome if the delegating handler used the same code-path.

barsh:
@joukevandermaas I have incorporated all of your feedback.

The delegating handler and `JsonApiAttribute` are now using the same method `JsonApiProcessor.ProcessRequest`, which means we might not need to duplicate the tests, but I think we do need to add the following test:

`JsonApiAttribute should respond with json api even when client does not request media type 'application/vnd.api+json'`

Can you give me some direction on where this test should go and point me towards an example that I can use as a reference?

joukevandermaas:
@barsh awesome! I will review your changes as soon as possible. I think [these tests](https://github.com/joukevandermaas/saule/blob/master/Tests/Integration/ContentNegotiationTests.cs#L106) have some setup that should allow you to write the test you propose. They create an http client that talks to some of the endpoints in an example application ([see here](https://github.com/joukevandermaas/saule/tree/master/Tests/Controllers)). You could add a controller or action method there.

Alternatively you could create a new file that just tests the new process method you created (whichever you find easier to do).

barsh:
@joukevandermaas see added test 0f1b7b2

joukevandermaas:
@barsh thanks so much for seeing this through! I'll publish a new pre-release with these changes.

Issue #201, opened 2018-09-24
Discuss new LinkType.Top


sergey-litvinov-work:
Hi Jouke,

I wanted to ask one thing and i can make PR to add it, but wanted to get your opinion about it. 

We have a case when we have endpoint that returns collection of items and we don't want to add additional endpoint to get specific item. The problem with the links as by default it will generate both `self` in each item and then `links->self` at top level. We want to be able to generate top level `self` but without `self` on each item level. 

Right now it's impossible as if i set `WithLinks(Saule.LinkType.Related);` in the resource description, then it wont generate both `self` links and we need that top one.

Is it ok to add `LinkType.Top` like this

```
public enum Test
{
        /// <summary>
        /// No links
        /// </summary>
        None = 0,

        /// <summary>
        /// Only self links
        /// </summary>
        Self = 1,

        /// <summary>
        /// Only related links
        /// </summary>
        Related = 2,

        /// <summary>
        /// Only top links
        /// </summary>
        Top = 4,

        /// <summary>
        /// Generate all possible links
        /// </summary>
        All = ~None
}
```

So then i can specify `LinkTypes.Top` and it will display only top `links.self` section without `self` for each item. 

And to prevent existing behavior i can check in the code (ResourceSerializer.CreateTopLevelLinks) to render top `self` link only if 

```
if (_resource.LinkType.HasFlag(LinkType.Top) && !_resource.LinkType.HasFlag(LinkType.Self)
 || _resource.LinkType.HasFlag(LinkType.Self))
{
      result.Add("self", _baseUrl);
}
```

So existing code will work fine as Self would be treated as Top too. And if someone doesnt want it, then he can specify it in the resource as `LinkType.Top & ~LinkType.Self`

joukevandermaas:
I think this makes sense, but I'm not sure I understand the reasoning for calling this `Top`. It sounds like it means "add the link for the collection, but not the items". Could we call it something like `CollectionSelf`? In any case I think it should have `Self` in the name. Does that make sense to you?

sergey-litvinov-work:
Totally makes sense. And as i renamed it to TopSelf and you merged it, i'm closing this defect. Thanks!

sergey-litvinov-work:
@joukevandermaas or if you think that CollectionSelf is better, i can rename it and make new PR

PR #202, opened 2018-09-24
added LinkType.Top enum, so we can specify only top level links

sergey-litvinov-work:
Hi Jouke,

It's a PR for #201 item. I tried to preserve back compatibility and all existing tests are passing fine. I also added new one to test Top only scenario.

Thanks

### DIFF
--- a/Saule/LinkType.cs
+++ b/Saule/LinkType.cs
@@ -24,9 +24,9 @@ namespace Saule
         Related = 2,
 
         /// <summary>
-        /// Only top section links
+        /// Only self links in the top section
         /// </summary>
-        Top = 4,
+        TopSelf = 4,
 
         /// <summary>
         /// Generate all possible links
--- a/Saule/LinkType.cs
+++ b/Saule/LinkType.cs
@@ -24,6 +24,11 @@ namespace Saule
         Related = 2,
 
         /// <summary>
+        /// Only top section links
+        /// </summary>
+        Top = 4,
+
+        /// <summary>
         /// Generate all possible links
         /// </summary>
         All = ~None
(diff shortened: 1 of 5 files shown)


Saule/Serialization/ResourceSerializer.cs
```diff
@@ -130,7 +130,10 @@ private JToken CreateTopLevelLinks(int count)
         {
             var result = new JObject();
 
-            if (_resource.LinkType.HasFlag(LinkType.Self))
+            // if resource has Top only and not Self, then we render it.
+            // otherwise to preserve back compatibility if Self is enabled, then we also render it
+            if ((_resource.LinkType.HasFlag(LinkType.Top) && !_resource.LinkType.HasFlag(LinkType.Self))
```
joukevandermaas: I think this can be simplified to `Top || Self`?

Saule/LinkType.cs
```diff
@@ -23,6 +23,11 @@ public enum LinkType
         /// </summary>
         Related = 2,
 
+        /// <summary>
+        /// Only top section links
+        /// </summary>
+        Top = 4,
```
joukevandermaas: Like I said on the issue, should this contain the word `Self` since it only applies to that?

Saule/Serialization/ResourceSerializer.cs
sergey-litvinov-work: Agree. Updated it

Saule/LinkType.cs
sergey-litvinov-work: Updated it. Now it's `TopSelf`

sergey-litvinov-work:
Thanks!

PR #204, opened 2018-10-02
Allow attributes on class

dejarp:
I'd like to be able to specify DisableDefaultIncludedAttribute, AllowsQueryAttribute, and PaginatedAttribute at the class (ApiController) level so that all endpoints in a controller and any derived controllers share these settings. This is especially important for DisableDefaultIncludedAttribute, because accidentally including all related resources can have unknown performance implications.


joukevandermaas:
Makes sense! thanks!

PR #205, opened 2018-10-08
Added a new filter expression for multiple search on strings

PhyberApex:
Added tests for the new expression


PhyberApex:
So I did some tinkering around and am now at a point where I don't really know how to go from here. I think my experience with C# starts to 'shine'. I never worked with reflection. Any chance you could give me some hint how to fix this one? I am stuck [here](https://github.com/PhyberApex/saule/blob/feature/csvfilter/Saule/Queries/Lambda.cs#L25)

~Cheers

joukevandermaas:
@PhyberApex sorry for the late reply. The ultimate goal of this piece of code is to generate an `Expression` that looks like this: `thing.filterProp == filterValue`. The *exact* expression comes from the user though, they can register them on the `QueryFilterExpressionCollection` (as you know). So we need to somehow make it easy to write these things.

The secondary goal then is to create expressions for `thing.filterProp` and `filterValue` so we can pass them to the filter expression lambda registered by the user. As you know, they can create a lambda like `(prop, filter) => prop == filter`. However, when we actually run the `Where` query, *that* thing takes a lambda like `(val) => val == constant`. Basically, we need to convert the former to the latter. This is done by the method called [`Convert`](https://github.com/PhyberApex/saule/blob/feature/csvfilter/Saule/Queries/Lambda.cs#L55) on the `Lambda` class.

As you can see, that method is generic, because all the `Expression` apis are generic. This kind of sucks for us, because we don't know the types at compile time, only at runtime. This is why we need to use reflection.

The line you linked essentially does the following:

```
Lambda.Convert<type>(expression, parsedValues, propertyExpression, param);
```

However since `type` in this case is an instance of `Type` and not an actual generic type parameter, you cannot just 'call' that line, but have to create it.

First we get the reflection `Type` instance of the class we're in (`Lambda`) using `typeof`. Then we look through all methods on the class (again, this is the class we're actually in) to find our static, private method called `Convert`. Now this `MethodInfo` represents the method, but like we said it's a generic method. This means we need to provide the type parameter before we can call it (`MakeGenericMethod`).

After doing all that we can call `Invoke` on the method info, setting the instance to `null` (because `Lambda` is a static class).

Hopefully (if the above was clear), you can see that the `expression` variable in the method you linked, only filters on one value. It needs to be changed to do the `||` chaining, for each filter. The rest of the code can probably stay the same.

PhyberApex:
Just to make sure here. With 

> The rest of the code can probably stay the same.

You mean like it was before I started or how it is right now? Because currently I can't build the solution and I can't really figure out what to do to fix that. So I am not sure if you said that what I did so far is in the wrong direction or if I am going the right way. Because the splitting of the actual query parameter has to happen somewhere.

~Cheers

joukevandermaas:
I realize this line "everything else can stay the same" is super confusing :smile: sorry about that. What I meant was the code as it is *in that specific method* can stay the same, *as you have it*. The build fails because it says `new[] { ... }` but the types of the objects within `{ }` are different. So changing that to `new object[] { ... }` will fix the build\*. Basically those array elements map to the parameters of the `Convert` method below.

Down lower at `Convert`, the second parameter is now of type `TProperty` but this must become `List<TProperty>` to match what it is passed. The `curriedBody` here is the actual expression `prop == constant`, where `constant` is the value from the filter. So we must create a `curriedBody` for each of the values instead of just the one, then chain them together using `Expression.OrElse`. The result is a single expression, which we must pass where `curriedBody` went before.

I definitely think you're on the right track and I think it's all very close to working. This part of the code is certainly not the most straightforward. Thanks for doing this work!

\* *(There's also some build errors in the test project because you changed `Value` to `Values`. I guess if possible we should have both properties, so we don't break backwards compatibility. But this is really not that urgent and I wouldn't worry about it too much for the moment)*

PhyberApex:
Thank you for helping me understand what I am actually doing here and always being so responsive!

I will take a look at this soon!

~Cheers

PhyberApex:
Sorry for not trying again sooner. with your help I got somewhat further with this. I changed the parameter to object[] and added the expression chaining. It does compile now but I get a runtime error on the reflection part while running the tests that a list of object could not be cast to int/string. My guess would be that as the return value of TryConvert is now a list of objects instead of a single one and a conversion is not as easy now. I tried to get it done with some the .ofType<T> method on the object array but I guess that's not working as we do not know the type at compile timeso my guess would be  this would probably require reflection again? I'm not too sure tbh. I hope you can shed some light into this for me.

PS:
I did take a look at the backwards compatibility tho and fixed that because I got frustrated with the first part :)

~Cheers

PhyberApex:
Okay so apperantly it was just too late yesterday and I got this working now. Made some mistake with chaining of expression tho. Gonna take a look at that soonish.

~Cheers

PhyberApex:
So apparently it was just way to late yesterday for me and I got this working now. There are other errors now in the chaining of the expression will take a look at this soonish.

~Cheers

PhyberApex:
So apparently it was just way to late yesterday for me and I got this working now. There are other errors now in the chaining of the expression will take a look at this soonish.

~Cheers

PhyberApex:
So apparently it was just way to late yesterday for me and I got this working now. There are other errors now in the chaining of the expression will take a look at this soonish.

~Cheers

PhyberApex:
So apparently it was just way to late yesterday for me and I got this working now. There are other errors now in the chaining of the expression will take a look at this soonish.

~Cheers

PhyberApex:
So apparently it was just way to late yesterday for me and I got this working now. There are other errors now in the chaining of the expression will take a look at this soonish.

~Cheers

PhyberApex:
So apparently it was just way to late yesterday for me and I got this working now. There are other errors now in the chaining of the expression. Will take a look at this soonish.

~Cheers

Saule/Queries/Lambda.cs
```diff
@@ -50,15 +51,27 @@ internal static List<object> TryConvert(List<string> values, Type type)
...
         private static Expression<Func<TClass, bool>> Convert<TProperty, TClass>(
             Expression<Func<TProperty, TProperty, bool>> expression,
-            TProperty constant,
+            List<TProperty> constant,
             MemberExpression propertyExpression,
             ParameterExpression parameter)
         {
-            var curriedBody = new FilterLambdaVisitor<TProperty>(propertyExpression, constant).Visit(expression.Body);
+            // initialize the expression with a always true one to make chaining possible
+            Expression curriedBody = Expression.Lambda(Expression.Constant(true));
```
joukevandermaas: should be `false`? (`true || a == b || a == c || a == d` is always true)

Saule/Queries/Lambda.cs
```diff
@@ -50,15 +51,27 @@ internal static List<object> TryConvert(List<string> values, Type type)
             return parsedList;
         }
 
+        internal static object TryConvert(string value, Type type)
```
joukevandermaas: I think we need to call this for each filter, so instead of 

```csharp
var parsedValues = TryConvert(values, valueType);
```

(line 17)

```csharp
var parsedValues = values.Select(v => TryConvert(v, valueType)).ToList();
```

PhyberApex:
So apparently it was just way to late yesterday for me and I got this working now. There are other errors now in the chaining of the expression. Will take a look at this soonish.

~Cheers

Saule/Queries/Lambda.cs
PhyberApex: You are right my bad!

Saule/Queries/Lambda.cs
PhyberApex: I had two `TryConvert` methods. One which took an array and one that only took a single one. I have those reduced to one now and added the `Select` function at the appropriate places.

PhyberApex:
The github outtage is given me quite a hard time atm. Sorry for comment spamming!

I did add the changes like you suggested. Will take a look at the other stuff soon.

~Cheers

PhyberApex:
I think this is it. Please let me know if you think there need to be any addition tests or changes. Also if you could let me know how to run that code style check within Visual Studio that would help me out not having to commit to wait for AppVeyor to tell me whats wrong :)

~Cheers

PhyberApex:
As I tried to get the nested filter working for the other PR I discovered an error in my CSV regex if the value had whitespace. I added a unit test for this as well.

~Cheers

Saule/Queries/Filtering/FilterProperty.cs
```diff
@@ -9,11 +13,34 @@ public class FilterProperty
...
+        public FilterProperty(string name, string values)
         {
-            Value = value;
             Name = name.ToPascalCase();
+            Value = values;
+
+            // Spliting the string into multiple values with csv notation
+            Values = new List<string>();
+
+            var match = new Regex("(?:,\"|^\")(\"\"|[\\w\\W]*?)(?=\",|\"$)|(?:,(?!\")|^(?!\"))([^,]*?)(?=$|,)").Matches(values);
```
joukevandermaas: Can you maybe put a comment that explains this regex? Some time in the future we may want to remember how it works.

Tests/Queries/FilterInterpreterTests.cs
```diff
@@ -125,6 +151,34 @@ public void WorksOnStrings()
...
+            Assert.Equal(expected, result);
+        }
+
+        [Fact(DisplayName = "Applies filtering on strings with multiple values")]
+        public void WorksOnStringsMultiple()
+        {
+            var people = Get.People(100).ToList().AsQueryable();
+            var expected = people.Where(c => c.LastName == "Russel" || c.LastName == "Comma,Test").ToList();
+
+            var query = GetQuery(new[] { "LastName" }, new[] { "Russel,\"Comma,Test\"" });
```
joukevandermaas: Maybe we can also see what happens on some edge cases/malformed strings such as `"a"b` (no comma), `"ab,c` (no end quote), etc.

FWIW i'm not sure what *should* happen in this case.. I would be fine with a 400 although technically that would be a breaking change.

PhyberApex: Right now this is failing silently as anything not captured will be ignored. It's certainly possible to add a 400. I am not sure if this would be the best way tho.

joukevandermaas: I would vote that either

1. We throw an error for this (400 bad request)
2. We identify the filter as "malformed" and interpret it as a single filter value

I don't think we should let parts drop, throw 500, etc.

I think you're right that option 2 is probably better.

Saule/Queries/Filtering/FilterProperty.cs
PhyberApex: I tried my best to explain this one. I hope that it's enough.

Tests/Queries/FilterInterpreterTests.cs
PhyberApex: I added a few checks and one more unit test for this.

PhyberApex:
Btw. do you want me to help you with the documentation of the PRs I did as well? I discovered that you are using the github pages to host the official website for saule and the source files for that are in the doc folder. So if you are okay with that I'd more than happy to do so.

~Cheers

joukevandermaas:
@PhyberApex yes, if you have time that would be great. For this PR I don't know if we really need much documentation, since this is basically what the spec describes.

PR #206, opened 2018-10-09
used _baseUrl.AbsoluteUri instead of _baseUrl to avoid included port number in the self link

sergey-litvinov-work:
Hi Jouke,

This is small PR to fix one small issue. It happens for us only when we host it under IIS (Win10\WinServer 2016 and probably all others too).

The issue is that top self link always includes port even if port is standard like this:

```
    "links": {
        "self": "http://localhost:80/SauleExample/api/values"
    }
```

or like this
```
    "links": {
        "self": "https://localhost:443/SauleExample/api/values"
    }
```

It's still valid url as it can contain port, but it looks a bit weird. Somehow IIS always pushes port in the `Request.Url` and `_baseUrl.OriginalUrl` always has it, and when we pass Uri as object for serialization, then Newtonsoft.Json uses `.OriginalUrl` and it gives us 80\443 ports for regular http\https hosting.

I pushed change to use `_baseUri.AbsoluteUri` in the `ResourceSerializer` and also updated a couple of unit tests as they were using Uri from JObject but now it has just regular string. I also added unit test for `ResourceSerializer` to test this specific case. Also i verified that non standard ports still are displayed under IIS\IIS Express

Here is also screenshot of VisualStudio watch that shows current values during debug

sauleexample_debug

Full response example before fix looks like it:

```
{
    "data": [
        {
            "type": "person",
            "id": "1",
            "links": {
                "self": "http://localhost/SauleExample/api/people/1/"
            },
            "attributes": {
                "name": "some person",
                "age": 42
            },
            "relationships": {
                "job": {
                    "links": {
                        "self": "http://localhost/SauleExample/api/people/1/relationships/job/",
                        "related": "http://localhost/SauleExample/api/people/1/job/"
                    },
                    "data": {
                        "type": "company",
                        "id": "11"
                    }
                }
            }
        }
    ],
    "links": {
        "self": "http://localhost:80/SauleExample/api/values"
    },
    "included": [
        {
            "type": "company",
            "id": "11",
            "links": {
                "self": "http://localhost/SauleExample/api/companies/11/"
            },
            "attributes": {
                "name": "Some company",
                "location": "some location"
            }
        }
    ]
}
```

### DIFF
--- a/Saule/Serialization/ResourceSerializer.cs
+++ b/Saule/Serialization/ResourceSerializer.cs
@@ -132,7 +132,7 @@ namespace Saule.Serialization
 
             if (_resource.LinkType.HasFlag(LinkType.Self))
             {
-                result.Add("self", _baseUrl);
+                result.Add("self", _baseUrl.AbsoluteUri);
             }
 
             var queryStrings = new PaginationQuery(_paginationContext);
--- a/Tests/Serialization/ResourceSerializerTests.cs
+++ b/Tests/Serialization/ResourceSerializerTests.cs
@@ -115,14 +115,14 @@ namespace Tests.Serialization
             var job = result["data"]["relationships"]["job"]["links"];
             var friends = result["data"]["relationships"]["friends"]["links"];
 
-            var self = result["links"]["self"].Value<Uri>().AbsolutePath;
+            var self = result["links"]["self"].Value<string>();
             var jobSelf = job["self"].Value<Uri>().AbsolutePath;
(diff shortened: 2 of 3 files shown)


sergey-litvinov-work:
@joukevandermaas just wanted to double check, you want me to do something more there or you just waiting for something? as you haven't merged these changes  :)

joukevandermaas:
@sergey-litvinov-work my apologies, I got distracted by other things and completely forgot.

sergey-litvinov-work:
Thanks!

PR #213, opened 2018-11-26
Moved QueryContext action parameter initialization to more common place

sergey-litvinov-work:
Hi Jouke,

We had a case when WebApi action has `ObjectContext` parameter and then `HandlesQuery` attribute passes current `ObjectContext` value as paramter. So endpoint can consume\handle current context.

I did two changes there:
1. I moved that code that initialize `ObjectContext` in action's parameter to more common place `QueryContextUtils` so now it will be available to any action that needs it, and not only by action that has `HandlesQuery`. So if endpoint with `AllowsQuery` wants to get current context for logging\other reasons, then it can get it

2. I also added `Fieldset` initialization to `HandlesQuery` so now it also can use `fields` logic too. As before that it would ignore it and `ObjectContext` won't parse actual `Fieldset` value

### DIFF
--- a/Saule/Http/HandlesQueryAttribute.cs
+++ b/Saule/Http/HandlesQueryAttribute.cs
@@ -6,6 +6,7 @@ using System.Web.Http;
 using System.Web.Http.Controllers;
 using System.Web.Http.Filters;
 using Saule.Queries;
+using Saule.Queries.Fieldset;
 using Saule.Queries.Filtering;
 using Saule.Queries.Including;
 using Saule.Queries.Sorting;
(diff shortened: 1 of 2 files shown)


joukevandermaas:
Thanks, this looks quite useful!

Just to make sure, this is *not* a breaking change, correct? It seems like it is not, but I'd rather make sure.

Also, do you think it is possible to write a test for this new functionality? I'd love not to break it accidentally in the future.

sergey-litvinov-work:
@joukevandermaas sorry for so long answer. I missed your answer. so it didn't work for us so well, so we handled it in different way on top of `Saule`. It shouldn't be breaking changes as initially it worked only for `HandlesAttribute` and after that change it will be working for every case

If you want this change, i can add more unit tests and update PR

PR #217, opened 2019-03-16
Selective dependency upgrade

bjornharrtell:
Alternative to https://github.com/joukevandermaas/saule/pull/216 without upgrading `Humanizer.Core` which appears to be pulling in alot of new dependencies.


bjornharrtell:
Further details on the change in `Humanizer.Core` at https://github.com/Humanizr/Humanizer/issues/798.

Issue #215, opened 2019-03-16
Support JSON API bulk extension


bjornharrtell:
See http://springbot.github.io/json-api/extensions/bulk/.

bjornharrtell:
I suppose this would require a custom route attribute that can handle different content types on the same path.

PR #218, opened 2019-03-16
Initial implementation to support bulk extension

bjornharrtell:
This is an initial spike to implement https://github.com/joukevandermaas/saule/issues/215.

### DIFF
--- a/Saule/Constants.cs
+++ b/Saule/Constants.cs
@@ -3,6 +3,7 @@
     internal static class Constants
     {
         public const string MediaType = "application/vnd.api+json";
+        public const string MediaTypeBulkExtension = "application/vnd.api+json; ext=bulk";
 
         public static class PropertyNames
         {
--- a/Saule/Http/BulkExtRouteAttributeAttribute.cs
+++ b/Saule/Http/BulkExtRouteAttributeAttribute.cs
@@ -1,0 +1,71 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Web.Http.Routing;
+
+namespace Saule.Http
+{
+    public class BulkExtRouteAttributeAttribute : RouteFactoryAttribute
+    {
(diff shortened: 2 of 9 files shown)


bjornharrtell:
@joukevandermaas do you think something like this could be a reasonable approach to support the bulk extension?

joukevandermaas:
@bjornharrtell looks like a good approach.

It's not very clear from the examples on the page you linked, but it seems that you should be able to post/patch/delete resources of different types in a single request?

That doesn't seem feasible with this solution, because the enumerable has a fixed type in the action method's parameter. Perhaps we could create a type like this:

```cs
class JsonApiType
{
    string Type { get; }
    T ConvertTo<T>();
}
```

Then force action methods with `BulkExtRouteAttributeAttribute` to take `IEnumerable<JsonApiType>`? Or perhaps we don't force it, but document that if you want to support multiple types, you'd do it this way?

bjornharrtell:
I don't think it makes sense to support different types, as the bulk extension works explicitly on a specific resource collection?

Unrelated, I note that there is a requirement for extension negotiation described at http://springbot.github.io/json-api/extensions/#extension-negotiation.

bjornharrtell:
Seems I jumped the gun here and that heterogeneous resource collections are indeed supported by JSON-API (although that's a bit of a surprise to me..), at least that's my interpretation of the discussion at https://github.com/json-api/json-api/issues/278. However if Saule currently assumes homogenous resource collections (which I think it does?) I think it's reasonable to have the same limitation here.

bjornharrtell:
And yet another realization is that this and other extensions have been deprecated (https://jsonapi.org/extensions/#prior-extensions). However, it still does seem quite useful anyway and it doesn't look like the upcoming JSON API 1.1 will include what seemed to be the intended replacement (Operations, https://github.com/json-api/json-api/pull/1254).

PR #223, opened 2019-10-16
Added IPagedResult and handling of it, so we can specify TotalItemsCount

sergey-litvinov-work:
Hi Jouke,

We have one case that we need to handle. we need to display correct `links.last` reference, so api consumer would understand how many pages. The problem is that we don't use `IQueryable` as we have some custom ORM

This PR is a try to deal with it. I did
1. Added `.editorconfig` with spaces as default indentation style, so VS will apply this settings right away and a developer won't need to change them in settings
2. Added `IPagedResult` interface with `TotalResultsCount` and `Items` properties and sample implemnetation `PagedResult<T> : IPagedResult` that uses `IList<T> inside`

The intention of `IPagedResult` is to say Saule how many records we have, and then during processing if item that was returned in the endpoint inherits that interface, then we use `TotalResultsCount` to calculate `links.last`. And also as we don't need to serialize `IPagedResults` in json, then code unwraps `pagedResult.Items` and passes it to serialization code. 

## FirstPage number (0\1)

There will be another PR that i want to mention. Saule uses `0` as the first page in paging. But we are using `1` as the first page. I want to ask about two ways that i'm thinking of and what is the better way for you
1. Just add another property `JsonApiConfiguration.Paging_FirstPage` with default value `0` but with ability to customize it. And then update logic to use value from config
2. We have `IUrlPathBuilder` and i can create another `IPagingBuilder` that will have a method `PagingLinks CreatePaging(PaginationContext context, object value)` where `PagingLinks` will be a DTO class with 4 string properties (First\Last\Next\Prev). And update existing `PaginationQuery` to be default implementation of that class. And other consumers can customize it by having different first page, or some custom logic for that


joukevandermaas:
Thanks for the PR! I will try to review it sometime today.

Saule/Queries/Pagination/IPagedResult.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Collections;
+
+namespace Saule.Queries.Pagination
+{
+    /// <summary>
+    /// Paged result that can be used to tell Saule how many pages we have
+    /// </summary>
+    public interface IPagedResult
```
joukevandermaas: I don't really understand why this needs to be an interface. I'd rather it's just a class so we can extend it later without breaking implementations.

Saule/Queries/Pagination/IPagedResult.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        /// <summary>
+        /// Gets total count of items across all pages
+        /// </summary>
+        int TotalResultsCount { get; }
+
+        /// <summary>
+        /// Gets data items
+        /// </summary>
+        IEnumerable Items { get; }
```
joukevandermaas: Not a big fan of the untyped enumerable. It also blocks people from doing client-powered filtering on database level using `IQueryable`. I think we need to put some thought into all the use-cases here and make sure we handle them, I think.

Saule/Queries/Pagination/PagedResult.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+
+namespace Saule.Queries.Pagination
+{
+    /// <summary>
+    /// Simple implementation of IPagedResult
+    /// </summary>
+    /// <typeparam name="T">Type of collection item</typeparam>
+    public class PagedResult<T> : IPagedResult
```
joukevandermaas: Ah I see you added an implementation as well, do we really need the interface still in that case?

Saule/Queries/Pagination/PaginationQuery.cs
```diff
@@ -26,6 +31,12 @@ public PaginationQuery(PaginationContext context)
             PreviousPage = isNumber && page > 0
                 ? CreateQueryString(context.ClientFilters, page - 1)
                 : null;
+            if (value is IPagedResult && context.PerPage.GetValueOrDefault(0) != 0)
+            {
+                var paged = (IPagedResult)value;
+                var totalPages = (int)Math.Ceiling((double)paged.TotalResultsCount / context.PerPage.Value);
+                LastPage = CreateQueryString(context.ClientFilters, totalPages);
```
joukevandermaas: `LastPage` is `CreateQueryString`? Either the method name is wrong or the property has the wrong value, I think?

Saule/Serialization/ResourceSerializer.cs
```diff
@@ -32,11 +33,26 @@ internal class ResourceSerializer
             IncludeContext includeContext,
             FieldsetContext fieldsetContext,
             IPropertyNameConverter propertyNameConverter = null)
+            : this(value, value, type, baseUrl, urlBuilder, paginationContext, includeContext, fieldsetContext, propertyNameConverter)
+        {
+        }
+
+        public ResourceSerializer(
+            object value,
+            object unwrappedValue,
```
joukevandermaas: the name implies here that `unwrappedValue` is somehow derived from `value`. So could we pass just `value` here? or you think the unwrapping is expensive? I think i need some more context to understand this part.

Saule/Queries/Pagination/PagedResult.cs
sergey-litvinov-work: yeah, so i also like more generic implementation like this instead of untyped interface, but it gives some problems:
1. to apply filtering\sorting\paging for IQueryable\IEnumerable we need to have actual data. And if endpoint returns `PagedResult<T>` then we need to get `PagedResult<T>.Items` that is typed property. In case of interface we can easily get it as untyped. If it's typed one, then we need to use reflection or expression trees, to get actual `Items` as we don't know type in the code and can't get them. But it's doable and if you want, i can drop interface and update code to use `PagedResult<T>` instead of interface

2. we need to understand original endpoint response that can be `PagedResult<T>` as we use it for paging and understanding `LastPage` in `PaginationQuery`. And for actual sorting\paging\filtering, So we need to unwrap actual `PagedResults<T>.Items` to apply operations on it. And we need to have both original endpoint result (for paging) and unwrapped result (for actual page\sort\filter operations)

Saule/Serialization/ResourceSerializer.cs
sergey-litvinov-work: Commented this point above too. We need to have both of original value and unwrapped one. as if we have `PagedResult<T>` then we can't apply sort\page\filter operations on top of it, as we need to use actual items that were returned. And `PagedResult<T>` won't be returned in the response as it will be used like meta data to understand total items count and use it to generate `links` section

sergey-litvinov-work:
@joukevandermaas  Added answers to comments and could you please check `FirstPage number` section in the PR. We use `1` as first page instead of `0` in Saule and i want to make PR about that too to handle both cases and proposed two solutions for that and want to understand what do you prefer

Thanks!

Saule/Queries/Pagination/IPagedResult.cs
sergey-litvinov-work: I'll also add implementation for IQueryable case too

Saule/Queries/Pagination/PaginationQuery.cs
edward-rosado: Mabe it should be `GenerateLastPageParameters`

joukevandermaas: Yeah I was just commenting on the method name. `CreateQueryString` seems like a bigger thing than just the value for the last page. I'm also curious why `LastPage` etc are strings, but that's another issue and I think not something you introduced here.

Saule/Queries/Pagination/PagedResult.cs
joukevandermaas: I see the issue now. As far as I remember we don't support those filtering/sorting on regular `IEnumerable`, only the generic version (but I could be wrong). So if it's just about serialization, I think it won't matter that much?

Since all `IEnumerable<T>` also implement `IEnumerable` we can maybe have an internal property that drops the type parameter?

joukevandermaas:
@sergey-litvinov-work thanks for providing all the details. It is more clear now and I see the point of the untyped enumerables.

As for the first page number, I'm a little bit conflicted. On the one hand I think this `IPaginationBuilder` is better, since it is much more powerful and if we do it right, could also allow for custom schemes. On the other hand, it seems like it will require a lot of design and I don't want to necessarily block this PR on that work.

I think adding this property (`FirstPageNumber`) is a good solution for now, but let's put it on the `PaginatedAttribute` instead of the more general configuration. The `PaginatedAttribute` makes very clear assumptions about the type of pagination used, while it should be possible to refactor the rest of Saule to allow a completely custom one.

sergey-litvinov-work:
@joukevandermaas i pushed updates.

1. I've refactored logic and removed `IPagedResult` and now code uses `PagedResult<T>`
2. I also removed that `unwrappedObject` as i added `TotalResultsCount` to `PaginiationContext` with internal setter. 
3. Now if we return `PagedResult<T>` in the endpoint, then it will set `PagedResult<T>.TotalResultsCount` to `PaginationContext.TotalResultsCount` and then will use it in `PaginationQuery` for `links.last`.
4. If we return regular IQueryable\IEnumerable and `PaginationContext.TotalResultsCount` is null, then it will do `Count()` for value and set result to `TotalResultsCount` and use it for `links.last`.  This one might be important as it will affect existing endpoints too. Like if some endpoints already returns `IQueryable`, then new code will do `queryable.Count()` on top of it to get `links.last` so it will be plus 1 SQL query (in case of SQL Server) for existing endpoint
5. Logic above is used only if PaginationContext exists and endpoint has `PaginatedAttribute` on it
6. Also added `FirstPageNumber` to `PaginatedAttribute` with default `0` value and ability to specify any 0\positive value there

Also added a couple of unit tests to handle these new scenarios

Saule/Queries/Pagination/PagedResult.cs
sergey-litvinov-work: Refactored that and mentioned new details in the comment at bottom

Saule/Queries/Pagination/IPagedResult.cs
sergey-litvinov-work: Implemented handling for Count for IQueryable\IEnumerable (as we already did it for Skip\Take for both interfaces)

Saule/Queries/Pagination/PaginationQuery.cs
sergey-litvinov-work: yeah. i didn't touch that part as it was part of existing logic

joukevandermaas:
OK I will take a look at the changes.

For what it's worth on point 4: I avoided this on purpose for exactly the reason you describe, so maybe if `TotalResultsCount` is null, we just don't set the last page link? Of course a `COUNT` query is not that expensive, but people may still want to have an option to avoid it.

sergey-litvinov-work:
that makes sense. I'll push update that will disable it and then eventually we can bring it back or make it configurable

Thanks!

Saule/Queries/Pagination/PagedResultQuery.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        private static object UnwrapPagedResult(object obj)
+        {
+            return obj.GetValueOfProperty(nameof(PagedResult<object>.Items), true);
+        }
+
+        private static int GetTotalResultsCount(object obj)
+        {
+            return (int)obj.GetValueOfProperty(nameof(PagedResult<object>.TotalResultsCount));
```
joukevandermaas: Since this property is `int?` doesn't that mean you should cast to `int?` here?

sergey-litvinov-work: I made it as `public int TotalResultsCount { get; set; }`  as from my understanding if we want to return this type, then it should be specified here. It's one of the main purposes for that class

sergey-litvinov-work:
Pushed one more fix as i found an issue in my PR. When we have `FirstPageNumber` to be `1`, then `links.next` also should be updated and become `page[number]=2`.

Fixed that and updated unit test to test `links.next` too

sergey-litvinov-work:
Thanks!

Issue #198, opened 2018-09-06
Calling an interface that has [DisableDefaultIncluded] removes the "data" part of all relationships not listed in ?include


Bendilin25:
Hi, 
I have an interface that returns a class with at least two relationships and that interface is set to [DisableDefaultIncluded]. 

A normal call (GET http://localhost/processes) returns everything as it should. 
```
{
	"data": [{
		"type": "processes",
		"id": "61e41d54-08d0-4314-8117-1f8b3436d486",
		"links": {
			"self": "http://localhost/processes/61e41d54-08d0-4314-8117-1f8b3436d486/"
		},
		"attributes": {
			"name": "Process112",
			"description": "this is a description"
		},
		"relationships": {
			"owner": {
				"links": {
					"self": "http://localhost/processes/61e41d54-08d0-4314-8117-1f8b3436d486/relationships/owner/",
					"related": "http://localhost/processes/61e41d54-08d0-4314-8117-1f8b3436d486/owner/"
				},
				"data": {
					"type": "accounts",
					"id": "7395f63d-3fea-4b1f-884f-fcf19ce66c7a"
				}
			},
			"lastEditedBy": {
				"links": {
					"self": "http://localhost/processes/61e41d54-08d0-4314-8117-1f8b3436d486/relationships/last-edited-by/",
					"related": "http://localhost/processes/61e41d54-08d0-4314-8117-1f8b3436d486/last-edited-by/"
				},
				"data": {
					"type": "accounts",
					"id": "de758e91-618f-49c7-be22-67dc3eb23272"
				}
			}
		}
	}],
	"links": {
		"self": "http://localhost:80/processes"
	}
} 
```

However calling GET http://localhost/processes?include=owner results in the removal of the "data" attribute for the other relationship that was not in the ?included list, in the below example "lastEditedBy".

```
{
	"data": [{
		"type": "processes",
		"id": "61e41d54-08d0-4314-8117-1f8b3436d486",
		"links": {
			"self": "http://localhost/processes/61e41d54-08d0-4314-8117-1f8b3436d486/"
		},
		"attributes": {
			"name": "Process112",
			"description": "this is a description"
		},
		"relationships": {
			"owner": {
				"links": {
					"self": "http://localhost/processes/61e41d54-08d0-4314-8117-1f8b3436d486/relationships/owner/",
					"related": "http://localhost/processes/61e41d54-08d0-4314-8117-1f8b3436d486/owner/"
				},
				"data": {
[...]

PR #224, opened 2019-10-23
Includes

PhyberApex:
Fix for #198 
Also did a cleanup of the LINQ stuff

### DIFF
--- a/Saule/Serialization/ResourceSerializer.cs
+++ b/Saule/Serialization/ResourceSerializer.cs
@@ -1,9 +1,9 @@
 ﻿using System;
 using System.Collections;
-using System.Collections.Generic;
 using System.Linq;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using Saule.Queries;
 using Saule.Queries.Fieldset;
 using Saule.Queries.Including;
 using Saule.Queries.Pagination;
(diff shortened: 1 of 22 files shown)


Saule/Serialization/ResourceSerializer.cs
```diff
@@ -192,22 +190,13 @@ private JToken SerializeData(ResourceGraph graph)
...
                     return new JArray();
                 }
-                else
-                {
-                    return JArray.FromObject(tokens);
-                }
+                return JArray.FromObject(tokens);
             }
-            else
+            if (!tokens.Any())
```
joukevandermaas: I think this part is the real bug fix?

Saule/Serialization/ResourceSerializer.cs
```diff
@@ -377,43 +363,35 @@ private JObject SerializeRelationships(ResourceGraphNode node)
 
         private JToken SerializeRelationshipData(ResourceGraphNode node, ResourceGraphRelationship relationship)
         {
-            // short circuit if not included in graph
-            if (!relationship.Included)
```
joukevandermaas: Are you sure this part can go?

PhyberApex:
Will write some tests as soon as I get some free time for it and will check if it is safe to remove that part.

~Cheers

PhyberApex:
Hey I updated this branch as we need it internally as well...hope that's okay. I also wrote the test to check this works correctly. Hope that's all good.

~Cheers

PR #226, opened 2019-10-24
Fixed `links.prev` logic if first page number isn't zero

sergey-litvinov-work:
Hi Jouke,

We found one missing case in initial pagination request. When we have FirstPage as `1`, then we shouldn't generate `links.prev` if current page is `1`. Like in example below `prev` link shouldn't be present as first page is `1`

```
    "links": {
        "self": "https://localhost/WebService4/v5/internal/locations/marketplace?page[number]=1",
        "first": "https://localhost/WebService4/v5/internal/locations/marketplace?page[number]=1",
        "next": "https://localhost/WebService4/v5/internal/locations/marketplace?page[number]=2",
        "prev": "https://localhost/WebService4/v5/internal/locations/marketplace?page[number]=0",
        "last": "https://localhost/WebService4/v5/internal/locations/marketplace?page[number]=12"
    }
```
I added unit test to cover that scenario and sorry for initially missed case :) 

Thanks!


sergey-litvinov-work:
Thanks!

PR #227, opened 2019-10-25
Fixed pagination when endpoint uses custom PagedResult<T> type

sergey-litvinov-work:
Hey Jouke,

Sorry, but it's another PR for another missing thing :) we tried to use custom PagedResult<T> class. We use it like 

```
	public class XPagedResult<T>: PagedResult<T>
	{
		public int? TotalPages { get; set; }
	}
```

And then for JSON.API it won't display TotalPages, but for regular JSON it will display that TotalPages field too plus properties in the base class. 

The problem is my initial code in `PagedResuleQuery.IsPagedResult` was checking for specific type but it should also check the base classes too. I've fixed it in this PR and added unit test to cover that scenario too

Thanks!

### DIFF
--- a/Saule/Queries/Pagination/PagedResultQuery.cs
+++ b/Saule/Queries/Pagination/PagedResultQuery.cs
@@ -37,13 +37,20 @@ namespace Saule.Queries.Pagination
                 return false;
             }
 
-            if (!obj.GetType().IsGenericType
-                || obj.GetType().GetGenericTypeDefinition() != typeof(PagedResult<>))
+            // we need to check the whole hierarchy of classes
+            var type = obj.GetType();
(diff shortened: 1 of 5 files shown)


sergey-litvinov-work:
I don't think it's possible in the current approach as we are checking for base generic type `PagedResult<>` and current one would be some specific typed instance like `CustomPagedResult<string>` that can't be casted to base one. We can do cast to base only if we know about specific GenericType that was used to made inheritor, then it would work

Like here - https://dotnetfiddle.net/dTGKr8

so it's either manually check full hierarchy for typeless generic `PagedResult<>` or create `PagedResult` for specific generic type, and then use `type.IsAssignableFrom`

PR #231, opened 2019-11-21
Allow for self link generation for custom routes

DtaggartMB:
Reason

When doing a Post/Put for a single resource object using a custom route for example /location/{id}/rooms where a room was being created, the generated return self link would be generated as /locations/48/rooms/ rather then the expected /locations/48/rooms/{roomid}

POST <url>/locations/48/rooms/ 
```
{ 
    "data" : 
    { 
        "type" : "rooms", 
        "attributes" : 
        { 
            "name" : "___hxLnCfZU60", 
            "capacity" : 100, 
            "description" : "Generated by automation using the v5 endpoint" 
        } 
   } 
}
```
Return
```
{
  "data" : {
    "type" : "rooms",
    "id" : "12684",
    "attributes" : {
      "name" : "___hxLnCfZU60",
      "capacity" : 100,
      "description" : "Generated by automation using the v5 endpoint"
    }
  },
  "links" : {
    "self" : "<url>/locations/48/rooms/"
  }
}
```

What we did was look to see if the returned resource was a collection type, if it isn't it would look for a ID to try to generate that self link. We used the override-able method BuildCanonicalPath to generate the link. If using the default logic the link would generate as /rooms/{id} which is the standard way if you don't have it stacked within another route. Since we do have a custom route our override method produces the correct output.

Now if we made a request to a get a collection of rooms. /locations/48/rooms/ the logic will default the self to the original requesting url/locations/48/rooms/

### DIFF
--- a/Saule/Serialization/ResourceSerializer.cs
+++ b/Saule/Serialization/ResourceSerializer.cs
@@ -71,7 +71,8 @@ namespace Saule.Serialization
 
             var isCollection = _value.IsCollectionType();
             string id = null;
-            if (!isCollection){
+            if (!isCollection)
+            {
                 id = dataSection["id"]?.ToString();
             }
 
--- a/Saule/Serialization/ResourceSerializer.cs
+++ b/Saule/Serialization/ResourceSerializer.cs
@@ -69,7 +69,13 @@ namespace Saule.Serialization
                 ["data"] = dataSection
             };
 
-            var links = CreateTopLevelLinks(dataSection is JArray ? dataSection.Count() : 0);
+            var isCollection = _value.IsCollectionType();
+            string id = null;
+            if (!isCollection){
+                id = dataSection["id"]?.ToString();
+            }
+
+            var links = CreateTopLevelLinks(dataSection is JArray ? dataSection.Count() : 0, id);
 
             if (links.HasValues)
             {
@@ -126,14 +132,21 @@ namespace Saule.Serialization
             }
         }
 
-        private JToken CreateTopLevelLinks(int count)
+        private JToken CreateTopLevelLinks(int count, string id = null)
         {
             var result = new JObject();
 
             // to preserve back compatibility if Self is enabled, then we also render it. Or if TopSelf is enabled
             if (_resource.LinkType.HasFlag(LinkType.TopSelf) || _resource.LinkType.HasFlag(LinkType.Self))
             {
-                result.Add("self", _baseUrl.ToString());
+                if (id != null && !_baseUrl.AbsolutePath.EndsWith(id))
+                {
+                    AddUrl(result, "self", _urlBuilder.BuildCanonicalPath(_resource, id));
+                }
+                else
+                {
+                    result.Add("self", _baseUrl.ToString());
+                }
             }
 
             var queryStrings = new PaginationQuery(_paginationContext, _value);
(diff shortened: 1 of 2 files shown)


Saule/Serialization/ResourceSerializer.cs
```diff
@@ -126,14 +133,21 @@ private ResourceGraphPathSet IncludedGraphPathsFromContext(IncludeContext contex
...
             if (_resource.LinkType.HasFlag(LinkType.TopSelf) || _resource.LinkType.HasFlag(LinkType.Self))
             {
-                result.Add("self", _baseUrl.ToString());
+                if (id != null && !_baseUrl.AbsolutePath.EndsWith(id))
+                {
+                    AddUrl(result, "self", _urlBuilder.BuildCanonicalPath(_resource, id));
+                }
+                else
+                {
+                    result.Add("self", _baseUrl.ToString());
```
joukevandermaas: Do you think it is possible to call the `BuildCanonicalPath(ApiResource)` overload here? I think it makes sense to give that control in all cases, since it's needed at least *some* of the time.

Then also the second part of the condition can go (`!_baseUrl.AbsolutePath.EndsWith(id)`).

Of course query parameters should be handled correctly in all cases (I see you added a test for that anyway, so it should continue to pass).

DtaggartMB:
```
if (id != null)
{
    AddUrl(result, "self", _urlBuilder.BuildCanonicalPath(_resource, id));
}  
else
{
    AddUrl(result, "self", _urlBuilder.BuildCanonicalPath(_resource));
}
```

We could do that, the only thing is that the URLBuilder has that ```EnsureEndsWith("/");``` logic so it would break some of the test and might break people who are currently using the library where they expect the links with out the ending '/'.

For example

image

As for Queries, since its using the path builder they would be lost. Would need to add in additional logic to add them to the end which could get messy due to the ```AddUrl``` logic also doing ```EnsureEndsWith("/");``` at the end of the path as well.   

If that is the direction you think we should go with this would be how I would set it up.
```
if (_resource.LinkType.HasFlag(LinkType.TopSelf) || _resource.LinkType.HasFlag(LinkType.Self))
{
    var selfLink = id != null ? 
        _urlBuilder.BuildCanonicalPath(_resource, id) : 
        _urlBuilder.BuildCanonicalPath(_resource);
    AddUrl(result, "self", selfLink, _baseUrl.Query);
}
```
Then for the AddUrl method
```
private JObject AddUrl(JObject @object, string name, string path, string query = null)
{
    if (string.IsNullOrEmpty(path))
    {
        return @object;
    }

    var start = new Uri(_baseUrl.GetLeftPart(UriPartial.Authority).EnsureEndsWith("/"));
    var uriBuilder = new UriBuilder(start);
    uriBuilder.Path = path.EnsureEndsWith("/");
    uriBuilder.Query = query?.Trim('?');
    @object.Add(name, new Uri(uriBuilder.Uri.AbsoluteUri));

    return @object;
}
```

Something along this line.

joukevandermaas: Yeah, I can see how this could potentially break people. I worry that it is now unclear when the builder will be used. If someone was trying to hook into it, it would only work some of the time, and I think the scenario is not very clear unless you ran into it already. So from that perspective I think it's better to always use the builder.

I think there are problems with it too tho, for example if there are two links to a resource:

`/things/1`
`/objects/3/thing`

as it is, the `self` link will point to whatever path was used to get the resource. But if this is changed, it will become the canonical path instead (likely `/things/1` in this case). I think that's strictly speaking an improvement, but definitely a breaking change.

So I guess I agree with you, better not to change it, because it could break people in a non-obvious way.

DtaggartMB: To be fair, I actually like that approach to always use it. You are right, it is unclear when it is used and not used to start and because of that there are inconsistencies in the tests where is will expect the ending '/' and some test that don't. At this point I think people would have coded around it.

PR #232, opened 2019-11-26
Fixed links.last when there are no any items in response

sergey-litvinov-work:
Hi Jouke,

Looks like we missed one more case initially. When endpoint doesn't return any items, then `links.last` will have incorrect result. 

If we would use `1` as first page number, then the response would be

```
{
  "data" : [ ],
  "links" : {
    "self" : "https://some/v5/equipment/",
    "first" : "https://some/v5/equipment/?page[number]=1",
    "last" : "https://some/v5/equipment/?page[number]=0"
  }
}
```

and `last` has `page[number]=0` that isn't correct and should be the same as `first`.
And if we don't have custom `firstPage` and by default it's `0`, then response would be

```
{
  "data" : [ ],
  "links" : {
    "self" : "https://some/v5/equipment/",
    "first" : "https://some/v5/equipment/?page[number]=1",
    "last" : "https://some/v5/equipment/?page[number]=-1"
  }
}
```
This PR is the fix for that logic and also a couple of unit tests to cover this scenario

### DIFF
--- a/Saule/Queries/Pagination/PaginationQuery.cs
+++ b/Saule/Queries/Pagination/PaginationQuery.cs
@@ -35,7 +35,9 @@ namespace Saule.Queries.Pagination
             {
                 // we also should add firstPage as if it's not a zero, then total page count should be shifted based on firstPageNumber
                 // with 60 elements and 20 page size. if firstPage is 0, then lastPage should be 2. If firstPage is 1, then lastPage should be 3
-                var totalPages = (int)Math.Ceiling((double)context.TotalResultsCount / context.PerPage.Value) - 1 + context.FirstPageNumber;
+                int totalPages = context.TotalResultsCount == 0
+                    ? context.FirstPageNumber
+                    : (int)Math.Ceiling((double)context.TotalResultsCount / context.PerPage.Value) - 1 + context.FirstPageNumber;
                 LastPage = CreateQueryString(context.ClientFilters, totalPages);
             }
         }
--- a/Tests/Integration/JsonApiMediaTypeFormatterTests.cs
+++ b/Tests/Integration/JsonApiMediaTypeFormatterTests.cs
@@ -541,6 +541,42 @@ namespace Tests.Integration
             }
         }
 
+        [InlineData("api/companies/paged-result-10items", 0)]
+        [InlineData("api/companies/paged-result-10items-first-page", 1)]
+        [Theory(DisplayName = "Paged result calculates correct first and last links when there items for less then full page")]
+        public async Task PagedResultOnePageResult(string baseUrl, int firstPageNumber)
+        {
+            int itemsCount = 10;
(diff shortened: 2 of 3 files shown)


Tests/Integration/JsonApiMediaTypeFormatterTests.cs
```diff
@@ -506,6 +506,41 @@ public async Task PagedResult(object baseUrl, int firstPageNumber)
...
+                foreach (var endpoint in endpoints)
+                {
+                    var result = await client.GetFullJsonResponseAsync($"{baseUrl}");
+                    var resultCount = ((JArray)result.Content["data"])?.Count;
+                    var last = result.Content["links"]["last"].Value<string>();
+                    var first = result.Content["links"]["first"].Value<string>();
+                    Assert.Equal(HttpStatusCode.OK, result.StatusCode);
+                    Assert.Equal(0, resultCount);
+                    Assert.EndsWith($"{baseUrl}?page[number]={firstPageNumber}", first);
+                    Assert.EndsWith($"{baseUrl}?page[number]={firstPageNumber}", last);
```
joukevandermaas: Maybe it makes sense to assert that first and last are the same? In at these scenarios:

1. there are fewer results than the page size
2. there are no results (special case)

they should be the same regardless of whether you start with 0 or 1.

Saule/Queries/Pagination/PaginationQuery.cs
```diff
@@ -35,7 +35,9 @@ public PaginationQuery(PaginationContext context, object value)
             {
                 // we also should add firstPage as if it's not a zero, then total page count should be shifted based on firstPageNumber
                 // with 60 elements and 20 page size. if firstPage is 0, then lastPage should be 2. If firstPage is 1, then lastPage should be 3
-                var totalPages = (int)Math.Ceiling((double)context.TotalResultsCount / context.PerPage.Value) - 1 + context.FirstPageNumber;
+                int totalPages = context.TotalResultsCount == 0
```
joukevandermaas: What about `context.TotalResultsCount <= context.PerPage.Value`?

Tests/Integration/JsonApiMediaTypeFormatterTests.cs
sergey-litvinov-work: Added one more unit tests and two endpoints that returns 10 items with PageSize default as 20, so it would be less than the whole page. So in that case last\first also would be the same and next would be missing

Saule/Queries/Pagination/PaginationQuery.cs
sergey-litvinov-work: It should follow existing logic and totalPages would be `1`. So `0` is just edge case that needed to be handled

sergey-litvinov-work:
Thanks!

PR #233, opened 2019-11-26
Fix self generation for custom route with text IDs

DtaggartMB:
Found issue due to the change made here 

https://github.com/joukevandermaas/saule/pull/231

Due to using the `EndsWith` method, I missed the case where a user could be using a text based ID with different casing. While developing a new endpoint I saw that the ID was appended twice on the end of the URL.

Added Unit test to catch this in the future.

Solution

Added the `StringComparison.InvariantCultureIgnoreCase` argument to the endswith logic.


joukevandermaas:
@DtaggartMB Do you think you can rebase this branch on master? There are a lot of commits now even though the diff is fairly small. I think this is not what you intended?

DtaggartMB:
Yeah I can

PR #234, opened 2019-11-27
Fix Self Links ID casing

DtaggartMB:
Found issue due to the change made here

#231

Due to using the `EndsWith` method, I missed the case where a user could be using a text based ID with different casing. While developing a new endpoint I saw that the ID was appended twice on the end of the URL.

Added Unit test to catch this in the future.

Solution

Added the `StringComparison.InvariantCultureIgnoreCase` argument to the `endswith` logic.


joukevandermaas:
Thanks!!

Issue #235, opened 2019-11-29
Sparse fieldset only works on attributes


PhyberApex:
The json:api states that sparse fieldsets should work on all fields...and fields include the relationships. This does currently not work in saule. I will make a PR for this very soon.

PR #236, opened 2019-11-29
Fix relatioship fieldset

PhyberApex:
This is the fix for #235 

As mentioned there relationships should also be filtered by the sparse fieldset.

### DIFF
--- a/Tests/JsonApiSerializerTests.cs
+++ b/Tests/JsonApiSerializerTests.cs
@@ -68,7 +68,7 @@ namespace Tests
 
         }
 
-        [Fact(DisplayName = "Uses query fieldset expressions if specified")]
+        [Fact(DisplayName = "Uses query fieldset expressions on attributes if specified")]
         public void UsesQueryFieldsetExpressions()
         {
             var target = new JsonApiSerializer<CompanyResource>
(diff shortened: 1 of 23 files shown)


PhyberApex:
Sorry for all those commits in this again... we were using this internally and we needed the changes from both PRs and I did not know how to handle this better.

~Cheers

PhyberApex:
Any update on this?

joukevandermaas:
I'm sorry, this completely slipped past me during the holidays. I will take a look at this very soon.

Tests/JsonApiSerializerTests.cs
```diff
@@ -83,6 +83,22 @@ public void UsesQueryFieldsetExpressions()
             Assert.NotNull(result["data"][0]["attributes"]["location"]);
             Assert.Null(result["data"][0]["attributes"]["number-of-employees"]);
         }
+        
+        [Fact(DisplayName = "Uses query fieldset expressions also on relationships if specified")]
```
joukevandermaas: I guess the "also" will lose its meaning pretty soon after this is merged :smile:

Tests/JsonApiSerializerTests.cs
```diff
@@ -83,6 +83,22 @@ public void UsesQueryFieldsetExpressions()
...
+            {
+                AllowQuery = true
+            };
+            var people = Get.People(1).ToList();
+            var result = target.Serialize(people, new Uri(DefaultUrl, "?fields[person]=Job"));
+            _output.WriteLine(result.ToString());
+
+            var relationships = result["data"][0]["relationships"];
+
+            Assert.Null(relationships["family-members"]);
```
joukevandermaas: Should you also assert that `Job` actually is there?

Tests/JsonApiSerializerTests.cs
PhyberApex: I do not quite get what you mean here, I hope I resolved the issue.

Tests/JsonApiSerializerTests.cs
PhyberApex: I should and I will :)

joukevandermaas:
Thanks very much!

Tests/JsonApiSerializerTests.cs
joukevandermaas: You did, it wasn't really an issue just a stylistic thing :smile: if you read the test title by itself, the "also" didn't make sense, it only did in the context of the other tests. It was a very minor thing anyway.

PhyberApex: Ahh I see. Didn't get it probably because english istn't my first language :)

Issue #237, opened 2020-01-13
Property specific JsonConverters


wernst:
Using Saule 1.8, EF 6.1.3

According to the documentation here - http://awesan.com/saule/content/4-customizing-serialization.html custom serialization can be achieved with the use of a `JsonConverter`. The example shown registers a global converter to the serializer. I have a use case where I would like to use a custom converter specifically on a few properties. Newtonsoft.Json provides a `JsonConverterAttribute` to achieve this (see https://www.newtonsoft.com/json/help/html/JsonConverterAttributeProperty.htm), but I have not been successful in getting Saule to call the `WriteJson` method of my custom converter (deserialization/`ReadJson` actually works fine). 

Is this expected and is there a way to customize serialization at the property level? Thanks!

sergey-litvinov-work:
Hi @joukevandermaas . We are also experiencing the same issue. We have a custom format for date times so we use `JsonConverter` attribute to handle specific properties in the different way, but it doesn't work as 

https://github.com/joukevandermaas/saule/blob/1f8b0a94e77b32126cf1fbacbf8626a209591a59/Saule/Serialization/ResourceSerializer.cs#L295-L299

that does

https://github.com/joukevandermaas/saule/blob/d578a65c6bdc15037bb6ba6437f661f342a7d330/Saule/ObjectExtensions.cs#L45-L49

and ignores them.

We can handle it in that method or in directly ObjectExtensions.GetValueOfProperty by checking for JsonConverterAttribute and using it if it's present and we can make PR with the actual change. But wanted to check with you, would that approach be fine with you and if not, do you have any ideas how that can be handled better? The need for us is to have a bit custom format for specific DateTime property

Thanks!

joukevandermaas:
I am a little wary of this, because there are tons of `JsonX` attributes and I don't want to set an expectation that we support them all. It looks like JSON converters run for each property, and you could have a conditional in there. So it should be possible to add support for any custom stuff from within the converter?

Alternatively I would be OK with adding a converter from within the resource, like

```csharp
var mySpecialConverter = new CustomJsonConverter();
Attribute("MyProperty", jsonConverter: mySpecialConverter);
```

At least then the relationship is clearly defined, and there is no confusion about which attributes are supported.

Would that work for your use-case?

sergey-litvinov-work:
[...]

PR #238, opened 2020-02-13
Add JsonConverter Functionality On Serialization

DtaggartMB:
Added the ability for JsonConverters attached to the sourceobject properties to be triggered. Had to add filtering and depth counter to the ContractResolver so we don't serialize the whole object.

After some discussion about how this could be added. We didn't want to try and add logic that was creating instances of the converters that we might have to support given the number of different converters available. This simply invokes the Converters through the standard Json.Net serialization process.

This is missing tests, I will add them if you think this is a good solution, but it passes all existing ones. This aims to solve issue #237 

Looking for input or alternative ways to achieve the same result.

### DIFF
--- a/Saule/Serialization/JsonApiContractResolver.cs
+++ b/Saule/Serialization/JsonApiContractResolver.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Serialization;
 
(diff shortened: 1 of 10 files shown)


Saule/Serialization/JsonApiContractResolver.cs
```diff
@@ -8,22 +9,47 @@ namespace Saule.Serialization
     internal class JsonApiContractResolver : DefaultContractResolver
```
DtaggartMB: Looking back at this, I think it might be better to create a separate Contractor for the source serializations instead of modifying the original.

edward-rosado:
This looks good to me. If i understand this correctly It should ensure backward compatibility and give the developer the option on what contract resolver to use.

Saule/Serialization/ResourceSerializer.cs
```diff
@@ -288,14 +292,16 @@ private JObject SerializeNode(ResourceGraphNode node, bool isCollection)
 
         private JObject SerializeAttributes(ResourceGraphNode node)
         {
+            var serializedSourceObject = JObject.FromObject(node.SourceObject, _sourceSerializer);
```
joukevandermaas: This is not something I think we can do, because now we serialize *everything* on the object, rather than just what is required for the JSON API response. That could potentially include properties that should not be serialized because they represent a big graph or a relationship.

I think one of the main reasons of switching to the `ResourceGraph` concept was to avoid that type of bug.

DtaggartMB: So inside the `SourceContractResolver` we do make sure to filter out the properties not assigned to the resource. Upon setting up the `sourceSerializer` we pass in the api resource to use.

joukevandermaas: Ah, I see it now. Do you think we can add a comment on this line and line 315 to say that only the relevant attributes as defined in the `ApiResource` class are serialized?

DtaggartMB: Yes of course, Ill also work on test for this all as well

joukevandermaas: Great, thank you so much for working on this!

DtaggartMB: Updated, np and thanks!

Issue #241, opened 2020-07-31
Polymorphic models support question


sergey-litvinov-work:
Hi Jouke, hope everything is good on your side. 

we are exploring a possibility to have endpoint that would return different types that inherits the same base abstract model class. For example
```
public abstract class MenuItem
{
	public int Id { get; set; }
}

public class ServiceMenuItem : MenuItem
{
	public string ServiceName { get; set; }
}

public class PackagesMenuItem : MenuItem
{
	public string PackageName { get; set; }
}
```

and then we could have 
```
[HttpGet]
[ReturnsResource(typeof(MenuItemApiResource))]
public IEnumerable<MenuItem> GetMenuItems()
{
...
```

And right now it doesn't work well, as `MenuItemApiResource` knows only about attributes of the base model so we can't return different types. The intention is to get something like this
```
{
    "data": [
        {
            "type": "service-menu-item",
            "id": "21",
            "links": {
                "self": "https://localhost:44300/WebService4/v5/customer/menu-items/21/"
            },
            "attributes": {
                "service-name": "some service"
            },
        },
        {
            "type": "packages-menu-item",
            "id": "42",
            "links": {
                "self": "https://localhost:44300/WebService4/v5/customer/menu-items/42/"
            },
            "attributes": {
                "package-name": "some package"
            }
        }
    ],
    "links": {
        "self": "https://localhost:44300/WebService4/v5/customer/menugroup/4"
    }
```

I can make a PR with the code change for that but wanted to get your opinion before that. We are thinking about two implementation approaches

## Logic in the ApiResource

We can update `ApiResource` to have following changes

```
private Dictionary<Type, Lazy<ApiResource>> _polymorphicTypes = new Dictionary<Type, Lazy<ApiResource>>();

public virtual void AddPolymorphicTypes<ModelType, ResourceType>()
    where ResourceType: ApiResource, new()
{
    if (!_polymorphicTypes.ContainsKey(typeof(ModelType)))
    {
        _polymorphicTypes.Add(typeof(ModelType), new Lazy<ApiResource>(() => new ResourceType()));
    }
}

public virtual ApiResource GetResourceByObject(object model)
{
    if (model == null)
        return this;

    return GetResourceByType(model.GetType());
}

public virtual ApiResource GetResourceByType(Type objectType)
{
    // if type is present, then we return it
    if (objectType != null && _polymorphicTypes.TryGetValue(objectType, out Lazy<ApiResource> apiResource))
        return apiResource.Value;

[...]

PR #243, opened 2020-08-17
Added IApiResourceProvider and IApiResourceProviderFactory so ApiReso…

sergey-litvinov-work:
Hi Jouke,

here is the change that i mentioned in #241 . basically i made

```
    public interface IApiResourceProviderFactory
    {
        IApiResourceProvider Create(HttpRequestMessage request);
    }

    public interface IApiResourceProvider
    {
        ApiResource Resolve(object dataObject);
    }
```

The first one called in `Saule/Http/PreprocessingDelegatingHandler.cs` and it creates a specific provider based on the current http request. And then it uses created `IApiResourceProvider` and passes it to `Saule/Serialization/ResourceGraph.cs` and `Saule/Serialization/ResourceSerializer.cs`. And they are calling it with current data object and it returns `ApiResource`. 

In the default implementation it uses the same `request.Properties.TryGetValue(Constants.PropertyNames.ResourceDescriptor, out object resource)` and then it always returns it. But potentially if consumer wants, it can have custom implementation that won't depend on `ReturnsResourceAttribute` and can make `ApiResource` based on own rules.

In our case we are creating specific resource based on the specific type.

I added 3 new unit tests `Tests/Integration/JsonApiFormatterInheritenceTests.cs` that validates it. And also as i changed `ApiResource` parameter to `IApiResourceProvider` in `ResourceSerializer` - i updated a bunch of unit tests to wrap existing `ApiResource` provider

I'll check it tomorrow to be sure that i didn't miss something but it should be ready for the review

### DIFF
--- a/Saule/Http/PreprocessingDelegatingHandler.cs
+++ b/Saule/Http/PreprocessingDelegatingHandler.cs
@@ -9,6 +9,7 @@ using System.Web.Http;
 using System.Web.Http.Controllers;
 using Saule.Queries;
 using Saule.Queries.Pagination;
+using Saule.Resources;
 using Saule.Serialization;
 
 namespace Saule.Http
@@ -43,11 +44,15 @@ namespace Saule.Http
 
             ApiResource resource = null;
             bool isHttpError = content is HttpError || content is IEnumerable<HttpError>;
-            if (request.Properties.ContainsKey(Constants.PropertyNames.ResourceDescriptor))
+            IApiResourceProvider resourceProvider = null;
+
+            if (!isHttpError)
             {
-                resource = (ApiResource)request.Properties[Constants.PropertyNames.ResourceDescriptor];
+                resourceProvider = config.ApiResourceProviderFactory.Create(request);
+                resource = resourceProvider?.Resolve(content);
             }
-            else if (content != null && !isHttpError)
+
+            if (resource == null && content != null && !isHttpError)
             {
                 content = new JsonApiException(
                     ErrorType.Server,
@@ -64,7 +69,7 @@ namespace Saule.Http
 
             PrepareUrlPathBuilder(jsonApi, request, config);
 
-            return jsonApi.PreprocessContent(content, resource, request.RequestUri, config);
+            return jsonApi.PreprocessContent(content, request.RequestUri, config, resourceProvider);
         }
 
         /// <inheritdoc/>
--- a/Saule/Http/PreprocessingDelegatingHandler.cs
+++ b/Saule/Http/PreprocessingDelegatingHandler.cs
@@ -49,7 +49,20 @@ namespace Saule.Http
             if (!isHttpError)
             {
                 resourceProvider = config.ApiResourceProviderFactory.Create(request);
-                resource = resourceProvider?.Resolve(content);
+                if (resourceProvider == null)
+                {
+                    content = new JsonApiException(
+                        ErrorType.Server,
+                        "ApiResourceProviderFactory returned null but it should always return an instance of IApiResourceProvider.")
+                    {
+                        HelpLink = "https://github.com/joukevandermaas/saule/wiki"
+                    };
+                    isHttpError = true;
+                }
+                else
+                {
+                    resource = resourceProvider.Resolve(content);
+                }
             }
 
             if (resource == null && content != null && !isHttpError)
--- a/Saule/JsonApiSerializer.cs
+++ b/Saule/JsonApiSerializer.cs
@@ -7,6 +7,7 @@ using Newtonsoft.Json.Linq;
 using Saule.Http;
 using Saule.Queries;
 using Saule.Queries.Pagination;
+using Saule.Resources;
 using Saule.Serialization;
 
 namespace Saule
@@ -32,7 +33,7 @@ namespace Saule
             return result.ResourceSerializer.Serialize(jsonSerializer);
         }
 
-        public PreprocessResult PreprocessContent(object @object, ApiResource resource, Uri requestUri, JsonApiConfiguration config)
+        public PreprocessResult PreprocessContent(object @object, Uri requestUri, JsonApiConfiguration config, IApiResourceProvider apiResourceProvider)
         {
             var result = new PreprocessResult
             {
@@ -61,6 +62,7 @@ namespace Saule
 
                 if (QueryContext != null && !QueryContext.IsHandledQuery)
                 {
+                    var resource = apiResourceProvider.Resolve(dataObject);
                     if (QueryContext.Filter != null)
                     {
                         dataObject = Query.ApplyFiltering(dataObject, QueryContext.Filter, resource);
@@ -79,7 +81,7 @@ namespace Saule
 
                 result.ResourceSerializer = new ResourceSerializer(
                         value: dataObject,
-                        type: resource,
+                        apiResourceProvider: apiResourceProvider,
                         baseUrl: requestUri,
                         propertyNameConverter: config.PropertyNameConverter,
                         urlBuilder: UrlPathBuilder,
--- a/Saule/JsonApiSerializerOfT.cs
+++ b/Saule/JsonApiSerializerOfT.cs
@@ -12,6 +12,7 @@ using Saule.Queries.Filtering;
 using Saule.Queries.Including;
 using Saule.Queries.Pagination;
 using Saule.Queries.Sorting;
+using Saule.Resources;
 using Saule.Serialization;
 
 namespace Saule
@@ -93,8 +94,9 @@ namespace Saule
             var queryContext = GetQueryContext(request.GetQueryNameValuePairs());
 
             _serializer.QueryContext = queryContext;
+            var apiResourceProvider = new DefaultApiResourceProvider(new T());
 
-            var preprocessResult = _serializer.PreprocessContent(@object, new T(), requestUri, config);
+            var preprocessResult = _serializer.PreprocessContent(@object, requestUri, config, apiResourceProvider);
             return JsonApiSerializer.Serialize(preprocessResult);
         }
 
--- a/Saule/JsonApiSerializerOfT.cs
+++ b/Saule/JsonApiSerializerOfT.cs
@@ -91,10 +91,11 @@ namespace Saule
             }
 
             var request = new HttpRequestMessage(HttpMethod.Get, requestUri);
+            ReturnsResourceAttribute.AddResourceToRequest(request, new T());
             var queryContext = GetQueryContext(request.GetQueryNameValuePairs());
 
             _serializer.QueryContext = queryContext;
-            var apiResourceProvider = new DefaultApiResourceProvider(new T());
+            var apiResourceProvider = config.ApiResourceProviderFactory.Create(request);
 
             var preprocessResult = _serializer.PreprocessContent(@object, requestUri, config, apiResourceProvider);
             return JsonApiSerializer.Serialize(preprocessResult);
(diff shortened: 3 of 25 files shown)


Saule/Http/PreprocessingDelegatingHandler.cs
```diff
@@ -43,11 +44,15 @@ public PreprocessingDelegatingHandler(JsonApiConfiguration config)
...
             ApiResource resource = null;
             bool isHttpError = content is HttpError || content is IEnumerable<HttpError>;
-            if (request.Properties.ContainsKey(Constants.PropertyNames.ResourceDescriptor))
+            IApiResourceProvider resourceProvider = null;
+
+            if (!isHttpError)
             {
-                resource = (ApiResource)request.Properties[Constants.PropertyNames.ResourceDescriptor];
+                resourceProvider = config.ApiResourceProviderFactory.Create(request);
+                resource = resourceProvider?.Resolve(content);
```
joukevandermaas, line 52: I think it's probably fair to throw if the factory returns null? I don't know if there's a scenario where that makes sense, and we can create a nice error message at this spot that allows people to figure out what went wrong.

Saule/JsonApiSerializer.cs
```diff
@@ -61,6 +62,7 @@ public PreprocessResult PreprocessContent(object @object, ApiResource resource,
 
                 if (QueryContext != null && !QueryContext.IsHandledQuery)
                 {
+                    var resource = apiResourceProvider.Resolve(dataObject);
```
joukevandermaas, line 65: is this needed? in that case, why did the serializer need to be changed?

Saule/JsonApiSerializerOfT.cs
```diff
@@ -93,8 +94,9 @@ public JToken Serialize(object @object, Uri requestUri, JsonApiConfiguration con
             var queryContext = GetQueryContext(request.GetQueryNameValuePairs());
 
             _serializer.QueryContext = queryContext;
+            var apiResourceProvider = new DefaultApiResourceProvider(new T());
```
joukevandermaas, line 97: Since this method gets a `JsonApiConfiguration` passed in, shouldn't it use the factory from there? I see why that's hard (no http request message), but it does make sense... Something to think about anyway.

Saule/Resources/DefaultApiResourceProviderFactory.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// </summary>
+    public class DefaultApiResourceProviderFactory : IApiResourceProviderFactory
+    {
+        /// <inheritdoc/>
+        public IApiResourceProvider Create(HttpRequestMessage request)
+        {
+            object resource;
+            if (request.Properties.TryGetValue(Constants.PropertyNames.ResourceDescriptor, out resource) && resource is ApiResource)
+            {
+                var apiResource = (ApiResource)resource;
```
joukevandermaas, line 16:
```suggestion
            if (request.Properties.TryGetValue(Constants.PropertyNames.ResourceDescriptor, out var resource) && resource is ApiResource apiResource)
```

Saule/Resources/DefaultApiResourceProviderFactory.cs
```diff
@@ -0,0 +1,24 @@
...
+                return new DefaultApiResourceProvider(apiResource);
+            }
+
+            // if no resource was specified then we return empty provider that will always return null
+            return new DefaultApiResourceProvider(null);
```
joukevandermaas, line 21: Does this match the current Saule behavior? I think it makes more sense to throw a nice error message that says what the problem is, but maybe this is not the spot for that?

Tests/Integration/JsonApiFormatterInheritenceTests.cs
```diff
@@ -0,0 +1,139 @@
...
+                _output.WriteLine(result.ToString());
+
+                var rectangle = result["data"];
+
+                ValidateRectangle(rectangle, "2");
+            }
+        }
+
+        
+        private static void ValidateRectangle(JToken rectangle2, string expectedId)
```
joukevandermaas, line 83: can just be `rectangle` now, probably this came from an extract method refactor?

Tests/Integration/JsonApiFormatterInheritenceTests.cs
```diff
@@ -0,0 +1,139 @@
...
+            Assert.Equal(expectedId, rectangle2["id"]);
+            Assert.Equal("Purple", rectangle2["attributes"]["color"]);
+            Assert.Equal(10, rectangle2["attributes"]["left"].Value<int>());
+            Assert.Equal(10, rectangle2["attributes"]["top"].Value<int>());
+            Assert.Equal(100, rectangle2["attributes"]["width"].Value<int>());
+            Assert.Equal(100, rectangle2["attributes"]["height"].Value<int>());
+            Assert.Equal(5, rectangle2["attributes"].Count());
+        }
+
+        private static void ValidateCircle(JToken circle1, string expectedId)
```
joukevandermaas, line 95: same with `circle`.

Tests/Integration/JsonApiFormatterInheritenceTests.cs
sergey-litvinov-work: yeah, it was refactoring and forgot to update variable name after extracting it. will fix

Saule/Resources/DefaultApiResourceProviderFactory.cs
sergey-litvinov-work: My initial code version had almost the same code - https://github.com/BookerSoftwareInc/saule/commit/0c166fd661a8d86014e059cc9e9f97d987c75391#diff-76813ea3f0bb77be87bd0ea4a4540059R13 but looks like AppVeyor uses older language version and it failed for me with compilation errors and i had to replace it to older syntax - https://ci.appveyor.com/project/JoukevanderMaas/saule/builds/34704891/messages

we can probably update AppVeyor to use 7.3 language version like csproj has and uses as locally that initial code compiles fine but i'm not totally sure how to do that

Saule/JsonApiSerializerOfT.cs
sergey-litvinov-work: that factory depends on `ReturnsResourceAttribute` that places `ApiResource` to `HttpRequestMessage.Properties`. but since we don't have it for these raw cases of `JsonApiSerializer` i went with creating it as is. I can also update the code to add `ApiResource` to `HttpRequestMessage.Properties` and then factory would work as well

Saule/JsonApiSerializer.cs
sergey-litvinov-work: Filtering\Paging\Sorting needs an instance of resource so we need to resolve it based on current `dataObject` and use it for these operations. Before that we had just plain instance of `ApiResource` and after the change we are resolving it on the fly

Saule/Http/PreprocessingDelegatingHandler.cs
sergey-litvinov-work: i had it initially to return null and fail in that case, but i found that there are unit tests like this - https://github.com/joukevandermaas/saule/blob/6d79f830f7442631e373ee581d702b01131dfb94/Tests/Controllers/CompaniesController.cs#L25-L31 . we don't have `ReturnsResource` here so no active `ApiResource` and code works fine as we don't have any `httpContent` as result and it's just `HttpStatusCode.NoContent`. 

from my understanding it makes sense to fail if it's return null as `ApiResource` always should be available, but then it would break that case when Endpoint doesn't return any object and doesn't need `ApiResource` so i updated it to allow null `ApiResource`

Saule/Resources/DefaultApiResourceProviderFactory.cs
sergey-litvinov-work: yeah, i described it above but there are some flows like `[HttpDelete]` where resource might be missing and `ApiResource` might be null in that case. and it just won't serialize anything as endpoint returns `void` that would be translated to Http 204 NoContent. if we would force `ApiResource` to always be present, then it might break some existing functionality

Saule/Http/PreprocessingDelegatingHandler.cs
joukevandermaas: Alright, I think that's a good reason. I don't want to break people's existing code that may depend on this.

Do you mind adding a comment to explain this?

Saule/JsonApiSerializer.cs
joukevandermaas: Is the data object never a collection then? I think the serializer now calls `Resolve` for each item in the collection but how does that work with sorting or filtering?

Saule/JsonApiSerializerOfT.cs
joukevandermaas: I think it would be confusing to ignore the provided factory, so I guess it would be fine to create a dummy http request message. It's confusing either way, but seems less confusing than to ignore the provided factory.

Saule/Resources/DefaultApiResourceProviderFactory.cs
joukevandermaas: I think I need to move to azure devops at some point, I'll do that eventually but for now it's fine to ignore this then.

Saule/JsonApiSerializer.cs
sergey-litvinov-work: correct, it might be collection. and then during the actual serialization it will be called for each item in the collection and might return different `ApiResource` based on the item. In the default implementation it always will be the same `ApiResource` that we have in `ReturnsResource`. 

so sorting\filtering still would work based on default `ApiResource`. if we are talking about our case where we might have different item types in the collection then filterting\sorting would work only for items that exist in base `ApiResource` as it might be very tricky to have different filtering\sorting depending on the item type in the same collection of base types

Saule/JsonApiSerializerOfT.cs
sergey-litvinov-work: ok, makes sense. i'll update it

Saule/Http/PreprocessingDelegatingHandler.cs
sergey-litvinov-work: yep, will do

Saule/JsonApiSerializerOfT.cs
sergey-litvinov-work: added a new internal method in ReturnsResource and used it in both places

Saule/Http/PreprocessingDelegatingHandler.cs
sergey-litvinov-work: so i checked it more and added a check for non null ApiResourceProvider. ApiResourceProvider itself might return a null ApiResource but it can't be null itself. Added a comment to the interface that mentions it. Also added an error if it's a null. And also added unit test to cover that scenario

Tests/Integration/JsonApiFormatterInheritenceTests.cs
sergey-litvinov-work: fixed that one and `circle` below as well

Saule/JsonApiSerializer.cs
sergey-litvinov-work: I added 3 more unit tests. Two of them are filtering and sorting by properties in the base `Shape` object and it works. Third one tries to sort by `Left` property in the inherited object `Rectangle` and fails as endpoint returns `IEnumerable<Shape>` and filtering can't find that property in the `Shape` object and throws expected `JsonApiException` with `Attribute 'left' not found.` title

joukevandermaas: Alright that makes sense, if you return a collection of multiple types, you can only sort/filter on properties they share. Nice that you added the tests for that 👍

joukevandermaas:
@sergey-litvinov-work do you think this is now ready to be merged? I don't see anything wrong with it, but not sure if you're working on it still.

sergey-litvinov-work:
@joukevandermaas yep, i think it's ready to be merged. i tested it for our case and it works great. if something would be missing i'll make another PR but for now it looks great
Thanks!

joukevandermaas (approved):
Thank you!

PR #244, opened 2020-08-20
Updated IApiResourceProvider to care about relationship\include cases

sergey-litvinov-work:
Hi Jouke, 

Looks like i missed one scenario related to relationship. without this change if we have a `Group` object that has `List<Shape> Shapes`, then it would return such response

```
{
    "data": [
        {
            "type": "group",
            "id": "1",
            "links": {
                "self": "http://localhost/api/groups/1/"
            },
            "attributes": {
                "name": "Group 1"
            },
            "relationships": {
                "shapes": {
                    "links": {
                        "self": "http://localhost/api/groups/1/relationships/shapes/",
                        "related": "http://localhost/api/groups/1/shapes/"
                    },
                    "data": [
                        {
                            "type": "shape",
                            "id": "1"
                        },
                        {
                            "type": "shape",
                            "id": "2"
                        },
                        {
                            "type": "shape",
                            "id": "3"
                        }
                    ]
                }
            }
        }
    ],
    "links": {
        "self": "http://localhost/api/groups"
    },
    "included": [
        {
            "type": "shape",
            "id": "1",
            "links": {
                "self": "http://localhost/api/shape/1/"
            },
            "attributes": {
                "color": "Purple"
            }
        },
        {
            "type": "shape",
            "id": "2",
            "links": {
                "self": "http://localhost/api/shape/2/"
            },
            "attributes": {
                "color": "Green"
            }
        },
        {
            "type": "shape",
            "id": "3",
            "links": {
                "self": "http://localhost/api/shape/3/"
            },
            "attributes": {
                "color": "Purple"
            }
        }
    ]
}
```

as it ignores that new `IApiResourceProvider` for that case. to handle it i added new method there
```
ApiResource ResolveRelationship(object dataObject, ApiResource relationship);
```
and now when code would serialize `data->relationship` section and `included` section, it will use it and pass `ApiResource` of the related resource itself. It happens here

https://github.com/joukevandermaas/saule/blob/c508f7bae2b73e0a32a29283d0c661a6a6eeefda/Saule/Serialization/ResourceGraph.cs#L128-L149

and we pass `r.RelatedResource` to recursive `Build` method and for these cases we will call that new method above and will pass that `r.RelatedResource` as second parameter for that method, so resolving can happen based on the relationship as it might depend on it. `Saule/Resources/DefaultApiResourceProvider.cs` would just return the same `r.RelatedResource` back but would allow to customize ApiResource based on the relationship

I also added two new unit tests (one that validates group.Shapes types) and another one that validates `included` section itself and types there.

After the change the output would be 
```
{
    "data": [
        {
            "type": "group",
            "id": "1",
            "links": {
                "self": "http://localhost/api/groups/1/"
            },
            "attributes": {
                "name": "Group 1"
            },
            "relationships": {
                "shapes": {
                    "links": {
                        "self": "http://localhost/api/groups/1/relationships/shapes/",
                        "related": "http://localhost/api/groups/1/shapes/"
                    },
                    "data": [
                        {
                            "type": "circle",
                            "id": "1"
                        },
                        {
                            "type": "rectangle",
                            "id": "2"
                        },
                        {
                            "type": "circle",
                            "id": "3"
                        }
                    ]
                }
            }
        }
    ],
    "links": {
        "self": "http://localhost/api/groups"
    },
    "included": [
        {
            "type": "circle",
            "id": "1",
            "links": {
                "self": "http://localhost/api/circles/1/"
            },
            "attributes": {
                "color": "Purple",
                "radius": 42
            }
        },
        {
            "type": "rectangle",
            "id": "2",
            "links": {
                "self": "http://localhost/api/rectangles/2/"
            },
            "attributes": {
                "color": "Green",
                "left": 10,
                "top": 10,
                "width": 100,
                "height": 100
            }
        },
        {
            "type": "circle",
            "id": "3",
            "links": {
                "self": "http://localhost/api/circles/3/"
            },
            "attributes": {
                "color": "Purple",
                "radius": 42
            }
        }
    ]
}
```

### DIFF
--- a/Saule/Resources/IApiResourceProvider.cs
+++ b/Saule/Resources/IApiResourceProvider.cs
@@ -11,5 +11,13 @@
         /// <param name="dataObject">Data object that is serialized. If it's an array, then it will be called for each item in the array</param>
         /// <returns>ApiResource that should be used for specific data object</returns>
         ApiResource Resolve(object dataObject);
+
+        /// <summary>
+        /// Returns ApiResource based on the data object and specific relationship
+        /// </summary>
+        /// <param name="dataObject">Data object that is serialized. If it's an array, then it will be called for each item in the array</param>
+        /// <param name="relationship">Relationship resource for which we are resolving api resource</param>
+        /// <returns>ApiResource that should be used for specific data object and relationship</returns>
+        ApiResource ResolveRelationship(object dataObject, ApiResource relationship);
     }
 }
--- a/Saule/Resources/DefaultApiResourceProvider.cs
+++ b/Saule/Resources/DefaultApiResourceProvider.cs
@@ -24,5 +24,11 @@
         {
             return ApiResource;
         }
+
+        /// <inheritdoc/>
+        public virtual ApiResource ResolveRelationship(object dataObject, ApiResource relationship)
+        {
+            return relationship;
+        }
     }
 }
--- a/Saule/Serialization/ResourceGraph.cs
+++ b/Saule/Serialization/ResourceGraph.cs
@@ -52,7 +52,7 @@ namespace Saule.Serialization
 
         private void Build(
             object obj,
-            ApiResource resource,
+            ApiResource relatedResource,
             ResourceGraphPathSet includePaths,
             int depth,
             string propertyName = null)
@@ -73,9 +73,17 @@ namespace Saule.Serialization
                 return;
             }
 
-            if (resource == null)
+            ApiResource resource;
+
+            // if we are serializing top object, then we just resolve it based on api resource of the endpoint
+            // but if are processing relationship's includes, then we need to resolve it based on the relationship itself
+            if (relatedResource == null)
             {
                 resource = _apiResourceProvider.Resolve(obj);
+            }
+            else
+            {
+                resource = _apiResourceProvider.ResolveRelationship(obj, relatedResource);
             }
 
             // keys (type & id pair) uniquely identifier each resource in a compount document
--- a/Saule/Serialization/ResourceSerializer.cs
+++ b/Saule/Serialization/ResourceSerializer.cs
@@ -424,7 +424,8 @@ namespace Saule.Serialization
 
                 foreach (var sourceObject in (IEnumerable)relationship.SourceObject ?? new JArray())
                 {
-                    content.Add(JObject.FromObject(new ResourceGraphNodeKey(sourceObject, relationship.Relationship.RelatedResource)));
+                    var resource = _apiResourceProvider.ResolveRelationship(sourceObject, relationship.Relationship.RelatedResource);
+                    content.Add(JObject.FromObject(new ResourceGraphNodeKey(sourceObject, resource)));
                 }
 
                 return content;
--- a/Tests/Controllers/ShapeController.cs
+++ b/Tests/Controllers/ShapeController.cs
@@ -6,13 +6,13 @@ using Tests.Models.Inheritance;
 
 namespace Tests.Controllers
 {
-    [ReturnsResource(typeof(ShapeResource))]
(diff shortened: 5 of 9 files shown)


Saule/Resources/IApiResourceProvider.cs
```diff
@@ -11,5 +11,13 @@ public interface IApiResourceProvider
...
         /// <returns>ApiResource that should be used for specific data object</returns>
         ApiResource Resolve(object dataObject);
+
+        /// <summary>
+        /// Returns ApiResource based on the data object and specific relationship
+        /// </summary>
+        /// <param name="dataObject">Data object that is serialized. If it's an array, then it will be called for each item in the array</param>
+        /// <param name="relationship">Relationship resource for which we are resolving api resource</param>
+        /// <returns>ApiResource that should be used for specific data object and relationship</returns>
+        ApiResource ResolveRelationship(object dataObject, ApiResource relationship);
```
joukevandermaas, line 21: Can you explain how the relationship can affect the type? As far as I can tell, the type should be the same whether the resource is used from within a relationship or not.

The case where it is not seems like an unlikely scenario to me, but at the same time the cost to having this extra method is very low, so maybe it doesn't matter.

joukevandermaas (commented):
Very nice PR as always :smile: I do have one small question, I wonder if you had a specific scenario in mind or if you just saw that it was cheap to provide extra information in case it was needed.

Saule/Resources/IApiResourceProvider.cs
sergey-litvinov-work: yeah, it's our specific case that we want to achieve. the idea is to have base type and inherited types. all of them would be available on the same endpoint but type and attributes might vary depending on the type. The closest example would be unit tests that i've added where we have a `Group` and it has `List<Shape> Shapes` that are marked with `HasMany<ShapeResource>`. And each Shape instance might be either Circle or Rectangle

with my previous's PR i made it possible to have endpoint that returns `public IEnumerable<Shape> GetAll()` and then each Shape might be different inherited type and have additional attributes compared to base Shape and it works great. 

But if we have a different resource like Group that has a reference to Shapes, then without this change - it still would return base `shape` type and attributes of the base class. And original change wasn't enough to handle that case as we have `GroupResource` that has `HasMany<ShapeResource>(..)` and we didn't call that initial `Resolve(object dataObject)` method for that case so it always used `ShapeResource`.  And to handle it i've added second Resolve that also would have relationship object itself so when we resolve `ApiResource`  we can do that based on specific `ApiResource relationship` and it might be different


in case of our specific scenario, we have some kind of online menu that has groups and some menu items inside and there are some properties that are the same for all of them and they are on the base menu item type. But then depending on the item itself properties might vary so we want to return different inherited type and additional attributes based on type so it gives us more flexibility

sergey-litvinov-work: just want to add that in the initial PR that `Resolve` method wasn't called at all for `included`\ `relationships` nodes and it always used `ApiResource` from `HasMany`\ `BelongsTo` and there wasn't any way to override it without the change. 

And it wasn't enough just to call `Resolve(..)` method for that case as default implementation always return original `ApiResource` from `ReturnsResource` attribute and it won't be the same as in `apiResource.Relationships[x].RelatedResource`. That second method helps as we just pass `RelatedResource` itself as parameter and then default implementation returns it back. And the same time allow us to override it in our implementation as well

joukevandermaas (approved):

Saule/Resources/IApiResourceProvider.cs
joukevandermaas: Right, so this is needed for the default implementation, not the custom ones. Presumably they would use reflection or something instead.

PR #245, opened 2020-10-28
Set encoding to UTF8 without BOM

developius:
`Encoding.UTF8` will include a UTF8 BOM character, which is invalid according to the [JSON spec RFC7159, Section 8.1](https://tools.ietf.org/html/rfc7159#section-8.1).
This was causing issues with (in particular) JavaScript clients that call `JSON.parse` on the response.
`UTF8Encoding` allows us to specify whether to include the BOM character or not.

### DIFF
--- a/Saule/Http/JsonApiMediaTypeFormatter.cs
+++ b/Saule/Http/JsonApiMediaTypeFormatter.cs
@@ -155,7 +155,7 @@ namespace Saule.Http
 
         private async Task WriteJsonToStream(JToken json, Stream stream)
         {
-            using (var writer = new StreamWriter(stream, Encoding.UTF8, 2048, true))
+            using (var writer = new StreamWriter(stream, new UTF8Encoding(encoderShouldEmitUTF8Identifier: false), 2048, true))
             {
                 await writer.WriteAsync(json.ToString(Formatting.None, _config.JsonConverters.ToArray()));
             }


Saule/Http/JsonApiMediaTypeFormatter.cs
```diff
@@ -155,10 +155,10 @@ public override bool CanWriteType(Type type)
 
         private async Task WriteJsonToStream(JToken json, Stream stream)
         {
-            using (var writer = new StreamWriter(stream, Encoding.UTF8, 2048, true))
+            using (var writer = new StreamWriter(stream, new UTF8Encoding(false), 2048, true))
```
joukevandermaas, line 158: Can you add a named parameter for this boolean? I think it will make the code just a little easier to understand.

joukevandermaas (commented):
Thanks for the PR! I have just a small suggestion.

Saule/Http/JsonApiMediaTypeFormatter.cs
developius: Good call. Done 😄

joukevandermaas (approved):
Perfect, thank you very much!

developius:
@joukevandermaas welcome 😁 

Looks like the release phase failed - what's involved in getting a release out for this?

joukevandermaas:
Ah it looks like my API key expired and API keys in general were deprecated 🤦 . I will have to dig into this, maybe I can move to github actions instead. I've been meaning to do that but never got around to it.

developius:
@joukevandermaas did you find any time to dig into GitHub actions?
We're using the patch in our project manually by copying the files instead of via Nuget, but it would be cool to get a release out to tidy things up for other people to use 😄 
Let me know if you would like any help, C# isn't my domain but maybe I can assist on the Actions side of things.

joukevandermaas:
@developius since it's the end of the year I've been pretty busy at work lately, so I didn't get to it yet. If I'm being honest I doubt that I'll get to it before the end of the year.

As far as I understand it should be fairly simple to set up github actions since we do barely anything (we literally just build the thing and publish the result). But Saule is still on the old build system (no `dotnet build` yet..) so maybe that needs to be fixed.

I'd be happy with a PR if it is possible to do that, I don't know enough about Github actions to know if it is.

PR #262, opened 2023-09-20
Pratham/pbis/1251070 correct failing build

ganeshlondhemindbodyonline:


joukevandermaas:
I think to make the build pass in CI we need to migrate away from Appveyor to Azure Devops. The publish to Nuget also no longer works because they changed it. I am no longer working on C# codebases so I will likely not have time to do this.

@ganeshlondhemindbodyonline If you or your team would like to maintain this repository I can add you as maintainers.

